PR #112, opened 2022-08-29
feat: 🎸 improve response time by using an SDK without tx queues

polymath-eric:
by using an SDK that uses batching the response times of some endpoints
is significantly improved. The most improved createAsset call can be up
to 4 times faster (1 minute -> 15 seconds)

### JIRA Link 
[DA-404](https://polymath.atlassian.net/browse/DA-404)

### Changelog / Description 

- Use an SDK version that eliminates transaction queues

### Checklist - 

- [ ] New Feature ?


src/transactions/transactions.service.ts
```diff
@@ -227,7 +227,7 @@ export class TransactionsService {
       };
 
       if (status === TransactionStatus.Succeeded) {
-        payload.result = 'placeholder'; // TODO @monitz87: use real result when we eliminate TQs
+        payload.result = transaction.receipt;
```
monitz87, line 230: This shouldn't be the receipt, but rather the result of running the transaction as resolved by the promise returned by `transaction.run`. The idea is for the subscriber to receive (for example) the id of the venue they just created, as if they had called the endpoint without subscribing.

This could be done with the current setup by assigning the result of `transaction.run` to a variable, then calling `catch` on it and then passing it to `assemblePayload` (which would have to be made async and might also involve some ugly type assertions).

My opinion though is that the `PolymeshTransactionBase` class in the SDK should be modified in one of two ways:

1. Adding a `public result?: TransformedReturnValue` property that is set before returning the result in the `run` method
2. Adding a `private _result?: TransformedReturnValue` property which is set in the same way as 1 and a `public get result(): TransformedReturnValue` getter that throws an error if called before the private variable has been set

I prefer 2 because it avoids clunky type assertions, and you can always avoid the error if you check that the transaction's status is `Succeeded` (by the way, the JSDoc for the getter should explain this)

monitz87 (changes requested):

monitz87:
Let it be known that this was the last PR reviewed by Jeremías Díaz Cordara for the Polymesh Association. Here I leave my mark, that future software archaeologists marvel at my thoroughness and precision

Take care guys, it was a fun ride

src/common/utils/index.ts
```diff
@@ -29,17 +30,17 @@ export type QueueResult<T> = {
...
-export async function processQueue<MethodArgs, ReturnType>(
-  method: WithArgsProcedureMethod<ProcedureMethod<MethodArgs, unknown, ReturnType>>,
+export async function processQueue<MethodArgs, ReturnType, TransformedReturnType = ReturnType>(
+  method: WithArgsProcedureMethod<ProcedureMethod<MethodArgs, ReturnType, TransformedReturnType>>,
   args: MethodArgs,
   opts: ProcedureOpts
-): Promise<QueueResult<ReturnType>> {
+): Promise<QueueResult<TransformedReturnType>> {
   try {
     const queue = await method(args, opts);
```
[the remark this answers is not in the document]
monitz87: Also `QueueResult` should be renamed

src/transactions/transactions.service.ts
```diff
@@ -227,7 +227,7 @@ export class TransactionsService {
       };
 
       if (status === TransactionStatus.Succeeded) {
-        payload.result = 'placeholder'; // TODO @monitz87: use real result when we eliminate TQs
+        payload.result = 'placeholder'; // The SDK needs to support returning this (DA-407)
```
polymath-eric, line 230: @VictorVicente I think this is OK to leave as a placeholder for now. 

I opened up DA-407 with Jere's comments to track the change needed.

VictorVicente: I'm ok with the placeholder too.

src/transactions/transactions.service.ts
polymath-eric: [DA-407](https://polymesh.atlassian.net/browse/DA-407)

VictorVicente (approved):

PR #114, opened 2022-09-06
feat: 🎸 allow webhookUrl param to be passed on write requests

polymath-eric:
a user can provide `webhookUrl` on any POST, PUT, PATCH or DELETE method. When given the API will not wait for block finalization (~15 seconds) before returning a response. The result of the transaction will be sent to the given webhook instead.

### JIRA Link 

✅ Closes: [DA-258](https://polymesh.atlassian.net/browse/DA-258)

### Changelog / Description 

- `webhookUrl` is now part of any endpoint that will create an on chain extrinsic. When passed the API will respond after performing the SDK validations and then post the result of the transaction to the given endpoint

### Checklist - 

- [X] New Feature ?


src/transactions/transactions.util.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+export type WithArgsProcedureMethod<T> = T extends NoArgsProcedureMethod<unknown, unknown>
+  ? never
+  : T;
+
+/**
+ * a helper function to handle when procedures have args and those without args
+ */
+export function prepareProcedure<MethodArgs, ReturnType, TransformedReturnType = ReturnType>(
```
polymath-eric, line 37: This is messy and untype safe, and is the main reason I put this up as a draft for now.

We did have some TODO comments about handling no args procedure throughout the code. I'll try to have another look at this to see if I can come up with something better.

The challenge is the overloaded signature where opts need to be passed in first when interacting with a NoArgs procedure

src/developer-testing/developer-testing.module.ts
```diff
@@ -0,0 +1,17 @@
+import { Module } from '@nestjs/common';
```
polymath-eric, line 1: I talked to Jere about this, and we were thinking this would be nice to have in to help testing out the feature.

I added in a flag so it won't show up in the docs normally.

src/transactions/config/transactions.config.ts
```diff
@@ -0,0 +1,11 @@
+/* istanbul ignore file */
+
+import { registerAs } from '@nestjs/config';
+
+export default registerAs('transactions', () => {
+  const { NOTIFICATIONS_LEGITIMACY_SECRET } = process.env;
```
polymath-eric, line 6: Even though this is needed here, I still prefixed with `NOTIFICATIONS` so it looks more consistent to the person setting them

src/transactions/transactions.util.ts
polymath-eric: The types are better now, albeit a bit verbose.

polymath-eric:
The code smells are from sonar cloud thinking we don't need a not null assertion, but TS says we do, and a TODO from Jere (now with my name).

For the duplication, its in a two test cases that are for accept / remove, so they look similar. I think its better to keep the parallel structure for them instead of refactoring since they are different methods so for them to share will need widely scoped variables.

prashantasdeveloper (changes requested):
Posting these comments for now. Will continue reviewing tomorrow

src/developer-testing/dto/webhook.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+export class WebHookDto {
+  @ApiProperty({
+    description: 'some text to send',
+    type: 'string',
+    example: '1',
+  })
+  @IsString()
+  readonly text: string;
+
+  constructor(dto: WebHookDto) {
```
prashantasdeveloper, line 15: I think we don't specify constructors for dtos

src/test-utils/service-mocks.ts
```diff
@@ -17,6 +19,14 @@ export class MockAssetService {
...
 }
 
+export class MockTransactionsService {
+  submit = jest.fn();
+}
+
+export const mockTransactionsProvider = {
+  provide: TransactionsService,
+  useValue: new MockTransactionsService(),
+};
```
prashantasdeveloper, line 29:
```suggestion
};

```

src/transactions/transactions.service.spec.ts
```diff
@@ -53,21 +79,63 @@ describe('TransactionsService', () => {
 
   afterEach(() => {
     mockIsPolymeshTransaction.mockReset();
+    mockIsPolymeshTransactionBatch.mockReset();
   });
 
   it('should be defined', () => {
     expect(service).toBeDefined();
   });
+  describe('submit (without webhookUrl)', () => {
```
prashantasdeveloper, line 88:
```suggestion
  
  describe('submit (without webhookUrl)', () => {
```

prashantasdeveloper (changes requested):

src/common/dto/signer.dto.ts
```diff
@@ -1,13 +1,19 @@
 /* istanbul ignore file */
 
-import { ApiProperty } from '@nestjs/swagger';
-import { IsString } from 'class-validator';
+import { ApiHideProperty, ApiProperty } from '@nestjs/swagger';
+import { IsOptional, IsString } from 'class-validator';
 
-export class SignerDto {
+export class TransactionBaseDto {
```
prashantasdeveloper, line 6: File name should be changed to `transaction-base.dto.ts`

src/common/decorators/swagger.ts
```diff
@@ -89,3 +93,21 @@ export const ApiPropertyOneOf = ({ union, ...apiPropertyOptions }: ApiPropertyOn
...
+ */
+export function ApiCreatedOrSubscriptionResponse(
+  options: ApiResponseOptions
+): ReturnType<typeof applyDecorators> {
+  return applyDecorators(
+    ApiCreatedResponse(options),
+    ApiAcceptedResponse({
+      description:
+        'Returned if `webhookUrl` is passed in the body. A response will be returned after the transaction has been validated. The result will be posted to the `webhookUrl` given when the transaction is completed',
+      type: NotificationPayloadModel,
```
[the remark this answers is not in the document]
polymath-eric: What does ApiExtraModels do? It seems like it's something to expose in the docs that isn't otherwise referenced directly.

It looks like NotificationPayloadModel shows up, so I think we don't need to. I am not entirely clear why its needed though.

https://docs.nestjs.com/openapi/types-and-parameters#extra-models

> HINT You only need to use @ApiExtraModels() once for a specific model class.

If we only need to call it once, then it should probably be used outside of a decorator thats called for every endpoint.

src/common/utils/index.ts
```diff
@@ -111,3 +18,24 @@ export function getTxTagsWithModuleNames(): string[] {
...
+
+export type ApiTransactionResponse = NotificationPayloadModel | TransactionQueueModel;
+
+// A helper type that lets a service return a QueueResult or a Subscription Receipt
+export type ServiceReturn<T> = Promise<
+  NotificationPayload<EventType.TransactionUpdate> | TransactionResult<T>
+>;
+
+// A helper type that lets a controller return a Model or a Subscription Receipt
+export type ModelResolver<T> = (res: TransactionResult<T>) => Promise<TransactionQueueModel>;
```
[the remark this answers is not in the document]
polymath-eric: `Transaction` make it seem like its an object, when really this type is a function that transforms a result to the model.

I am OK with something like `TransactionResolver<T>`, but I don't want it to be mistaken for an object when its meant to be a function.

src/developer-testing/dto/webhook.dto.ts
polymath-eric: We do it for 8 other ones.

look for `constructor` while filtering files to `**/*.dto.*`

I am OK removing it, I think I copied an example with it included

src/authorizations/authorizations.service.ts
```diff
@@ -55,19 +55,16 @@ export class AuthorizationsService {
...
   }
 
-  public async remove(id: BigNumber, signer: string): Promise<TransactionResult<void>> {
+  public async remove(id: BigNumber, signer: string, webhookUrl?: string): ServiceReturn<void> {
     const { remove } = await this.findOne(signer, id);
 
-    const address = await this.signingService.getAddressByHandle(signer);
-    // TODO: find a way of making processQueue type safe for NoArgsProcedureMethods
-    return processTransaction(remove, { signingAccount: address }, {});
+    return this.transactionsService.submit(remove, {}, { signer, webhookUrl });
```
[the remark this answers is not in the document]
polymath-eric: The old comment is out of date. (I missed one and was going to remove that as well)

This way will give us type checking for with excess property checking (type error for having a value in object).

I am handling the type unsafety in `prepareProcedure` in `transaction.utils` file.

src/common/dto/signer.dto.ts
```diff
@@ -1,13 +1,19 @@
...
     example: 'alice',
   })
   @IsString()
   readonly signer: string;
+
+  // Hide the property so the interactive examples work work without additional setup
+  @ApiHideProperty()
+  @IsOptional()
+  @IsString()
+  readonly webhookUrl?: string;
```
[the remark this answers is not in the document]
polymath-eric: I used @IsUrl(), which doesn't recognize `localhost` (which is why I didn't have it in the original PR).

I added a note in the readme to use either the IP, or make a /etc/host entry. Its possible to pass in a no tld option to the validator, but then its not as useful for real world use cases.

polymath-eric:
4 code smells on Sonar cloud to ignore. 1 TODO, 3 needed type assertions

prashantasdeveloper (approved):
LGTM (We can bump the SDK to v17)

PR #116, opened 2022-10-03
fix: 🐛 Modify authorizations accept/remove service logic

prashantasdeveloper:
### JIRA Link 

DA-369

### Changelog / Description 

Modifies logic for accept/remove service methods to correctly fetch authRequest for an id based on signer type

### Checklist - 

- [ ] New Feature ?


src/authorizations/authorizations.module.ts
```diff
@@ -2,14 +2,15 @@
...
+import { AccountsModule } from '~/accounts/accounts.module';
 import { AuthorizationsController } from '~/authorizations/authorizations.controller';
 import { AuthorizationsService } from '~/authorizations/authorizations.service';
 import { IdentitiesModule } from '~/identities/identities.module';
 import { PolymeshModule } from '~/polymesh/polymesh.module';
 import { TransactionsModule } from '~/transactions/transactions.module';
 
 @Module({
-  imports: [PolymeshModule, TransactionsModule, forwardRef(() => IdentitiesModule)],
+  imports: [PolymeshModule, TransactionsModule, forwardRef(() => IdentitiesModule), AccountsModule],
```
polymath-eric, line 13:
```suggestion
  imports: [PolymeshModule, TransactionsModule, AccountsModule, forwardRef(() => IdentitiesModule)],
```
(nit) I think it looks better having all forwardRef together

src/authorizations/authorizations.service.ts
```diff
@@ -55,14 +67,33 @@ export class AuthorizationsService {
     }
   }
 
+  private async getAuthRequest(address: string, id: BigNumber): Promise<AuthorizationRequest> {
+    const account = await this.accountsService.findOne(address);
+
+    const identity = await account.getIdentity();
+
+    let authRequest: AuthorizationRequest;
+    if (identity) {
```
polymath-eric, line 76: I am reading this like:

> If an Account has an Identity, then we get the identity auths, otherwise get the account auths.

Will this work? It seems like authorization can target an Identity or an Account.

Screen Shot 2022-10-03 at 2 37 59 PM

What about the case where an Authorization targets an Account that already has an Identity? Will falling back to the Identity lookup still work? I am not sure if the assumption this code is making will always work, although it might.

polymath-eric: I am thinking when there is an identity to call `this.findOne` two times and take either result.

It would be messier and I'd prefer if what you now is sufficient though.

polymath-eric (changes requested):

src/authorizations/authorizations.service.ts
prashantasdeveloper: I tried some of the authorizations and my logic seems to work. I too am not that happy with the logic here but couldn't figure out anything else. One other way that pops up in my mind is that we add a param specifying that the authorization being accepted is for identity/account, that way instead of guessing based on account identity we would proceed based on user input and if authorization is a mismatch anyhow that would throw an error. What do you think of this @VictorVicente ?

polymath-eric: I don't think we should ask users for input when we can double query to figure it out.

I think this will work for most cases and is better than the current implementation, but there might be a bug where an auth gets targeted to an Account but we only check the Identity. 

I can check to see if the storage will be transitive thorough the Account for the Identity

polymath-eric: I was able to reproduce the case I am thinking of. Join an Identity, and make an Auth targeting the account, this will never be able to accept it, even though its a valid auth.

Something like this seems to cover that case:
```ts
   const ignoreNotFound = (error: Error): undefined => {
      if (error instanceof NotFoundException) {
        return undefined;
      } else {
        throw error;
      }
    };

    let authRequest: AuthorizationRequest | undefined;
    if (identity) {
      authRequest = await this.findOne(identity, id).catch(ignoreNotFound);
    }

    if (!authRequest) {
      authRequest = await this.findOne(account, id);
    }

    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
    return authRequest!;
```

The exception is a bit annoying to deal with, espcially since its already a transformed error. Not sure on the cleanest way to handle it, but I think shoving it into a util function for catch isn't too bad.

src/transactions/transactions.service.ts
```diff
@@ -62,13 +62,17 @@ export class TransactionsService {
     this.legitimacySecret = config.legitimacySecret;
   }
 
+  public async getSigningAccount(signer: string): Promise<string> {
```
polymath-eric, line 65:
```suggestion
  private async getSigningAccount(signer: string): Promise<string> {
```
I like this method, but it should be private:

If someone needs the lookup they should import signingService directly.

polymath-eric (changes requested):

polymath-eric (approved):
LGTM, I looked at the code and why its public and it made more sense than seeing it isolated on Github

PR #118, opened 2022-10-11
feat: 🎸 Option to redeem tokens from a specific portfolio

prashantasdeveloper:
### JIRA Link 

DA-434

### Changelog / Description 

Provide option to redeem from a specific portfolio

BREAKING CHANGE: 🧨 `RedeemTokensDto` now contains `from` attribute to specify the portfolio from which tokens must be redeemed

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@nestjs/swagger": "^5.2.1",
     "@polymeshassociation/hashicorp-vault-signing-manager": "1.1.1",
     "@polymeshassociation/local-signing-manager": "1.2.0",
-    "@polymeshassociation/polymesh-sdk": "^17.0.0",
+    "@polymeshassociation/polymesh-sdk": "18.0.0-beta.2",
     "@polymeshassociation/signing-manager-types": "1.0.0",
     "class-transformer": "0.5.1",
     "class-validator": "^0.13.2",
--- a/src/assets/dto/redeem-tokens.dto.ts
+++ b/src/assets/dto/redeem-tokens.dto.ts
@@ -16,4 +16,14 @@ export class RedeemTokensDto extends TransactionBaseDto {
   @ToBigNumber()
   @IsBigNumber()
   readonly amount: BigNumber;
+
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "@nestjs/swagger": "^5.2.1",
     "@polymeshassociation/hashicorp-vault-signing-manager": "1.1.1",
     "@polymeshassociation/local-signing-manager": "1.2.0",
-    "@polymeshassociation/polymesh-sdk": "^17.0.0",
+    "@polymeshassociation/polymesh-sdk": "18.0.0-beta.2",
```
polymath-eric, line 36: We should do a `develop` -> `master` before this one get merged.

We should probably use semantic release bot in this repo and get alpha + beta branches as its own task. I made one [here](https://polymesh.atlassian.net/browse/DA-438)

src/assets/dto/redeem-tokens.dto.ts
```diff
@@ -16,4 +16,14 @@ export class RedeemTokensDto extends TransactionBaseDto {
   @ToBigNumber()
   @IsBigNumber()
   readonly amount: BigNumber;
+
+  @ApiProperty({
+    description:
+      'Portfolio number from which the Asset tokens must be redeemed. Use 0 for the Default Portfolio',
```
polymath-eric, line 22:
```suggestion
      'Portfolio id from which the Asset tokens must be redeemed from. Use 0 for the Default Portfolio',
```
I think id is more consistent than number.

polymath-eric (changes requested):
Code looks good, but we should merge current develop into master so we can still build on a 5.0.x version for now.

I'll open a PR for that

src/assets/dto/redeem-tokens.dto.ts
prashantasdeveloper: We are using `number` at all places.

polymath-eric (approved):

PR #121, opened 2022-10-14
feat: 🎸 add authentication mechanism

polymath-eric:
### JIRA Link 

[DA-259](https://polymesh.atlassian.net/browse/DA-259)

### Changelog / Description 

allows for api key based authentication by passing `x-api-key` as a request header. Configured by setting the ENV variable `AUTH_STRATEGY='apiKey'`

### Checklist - 

- [X] New Feature ?


src/auth/auth.service.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CreateApiKeyDto } from '~/auth/dto/create-api-key.dto';
+import { ApiKeyModel } from '~/auth/models/api-key.model';
+import { generateBase64Secret } from '~/common/utils';
+import { User, UserId } from '~/users/user.consts';
+
+@Injectable()
+export class AuthService {
+  private userStore: Record<UserId, User> = {};
+
+  constructor(config: ConfigService) {
```
polymath-eric, line 14: In other places we register our own dynamic config provider, but looking at tutorials it seems much more common to use the build in config service.

I think using the Nest service is better going forward and is much less boiler plate.

src/auth/strategies/strategies.conts.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const authStrategyValues = Object.values(AuthStrategy);
+
+/**
+ * A helper to be passed to `.sort`. The order is in which they will be evaluated.
+ *
+ * For example the the "open" strategy should always be last to allow for testing of other strategies
+ */
+export const cmpAuthStrategyOrder = (a: AuthStrategy, b: AuthStrategy): number =>
```
polymath-eric, line 17: I don't want someone to write `open,apiKey` config and wonder why apiKey never gets executed.

src/subscriptions/subscriptions.service.ts
```diff
@@ -262,15 +261,4 @@ export class SubscriptionsService {
 
     this.scheduleSendHandshake(id);
   }
-
-  // TODO @polymath-eric: implement random secret generation
```
polymath-eric, line 266: I think this was an inherited TODO. The crypto package should give sufficiently random bytes.

This is moved to util to make API keys

README.md
```diff
@@ -49,35 +76,37 @@ For any method that modifies chain state, the key to sign with can be controlled
...
-## Webhooks (alpha)
+### Authentication
+
+The REST API uses [passport.js](https://www.passportjs.org/) for authentication. This allows the service to be configurable with multiple strategies.
+
+Currently there are two strategies available:
+
+1. Api Key:
+   By configuring `apiKey` as a strategy, any request with the header `x-api-key` will be authenticated with this strategy. The env `API_KEYS` can be used to provide initial keys
+1. Open:
```
prashantasdeveloper, line 87:
```suggestion
2. Open:
```

src/auth/auth.service.ts
```diff
@@ -0,0 +1,61 @@
...
+import { parseApiKeysConfig } from '~/auth/auth.utils';
+import { CreateApiKeyDto } from '~/auth/dto/create-api-key.dto';
+import { DeleteApiKeyDto } from '~/auth/dto/delete-api-key.dto';
+import { ApiKeyModel } from '~/auth/models/api-key.model';
+import { generateBase64Secret } from '~/common/utils';
+import { User, UserId } from '~/users/user.consts';
+
+@Injectable()
+export class AuthService {
+  private userStore: Record<UserId, User> = {};
```
prashantasdeveloper, line 13: Any specific reason around `Record<UserId, User>`? Why not directly make it `Record<UserId, string[]>` to map api keys to userId?

src/auth/auth.service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * @param apiKey - key to delete
+   */
+  public async deleteApiKey({ apiKey }: DeleteApiKeyDto): Promise<void> {
+    const user = this.getUserByApiKey(apiKey);
+    if (user) {
+      user.apiKeys = user.apiKeys.filter(key => key !== apiKey);
+    } else {
+      throw new NotFoundException('api key was not found');
```
prashantasdeveloper, line 49:
```suggestion
      throw new NotFoundException('API key was not found');
```

prashantasdeveloper (changes requested):

README.md
polymath-eric: This is markdown so it increments on display. This way makes it easy to move points without messing up the numbering.
Screen Shot 2022-10-17 at 8 45 36 AM

README.md
```diff
@@ -36,48 +62,51 @@ SUBSCRIPTIONS_HANDSHAKE_RETRY_INTERVAL=## Amount of milliseconds between subscri
 NOTIFICATIONS_MAX_TRIES=## Amount of attempts to deliver a notification before it is considered failed ##
 NOTIFICATIONS_RETRY_INTERVAL=## Amount of milliseconds between notification delivery attempts ##
 NOTIFICATIONS_LEGITIMACY_SECRET=## A secret used to create HMAC signatures ##
+AUTH_STRATEGY=## list of comma separated auth strategies to use e.g. (`apiKey,open`) ##
```
sansan, line 65: Should list the env variable for API_KEYS

src/auth/auth.utils.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @throws if given invalid values
+ */
+export const parseAuthStrategyConfig = (rawStrategyConfig: string): AuthStrategy[] => {
+  const givenStrategies = rawStrategyConfig.split(',').map(strategy => strategy.trim());
+
+  const filteredStrategies = givenStrategies.filter(isStrategyKey);
+
+  if (filteredStrategies.length !== givenStrategies.length) {
+    throw new Error(
+      `Auth config error! "${givenStrategies}" contains an unrecognized option. Valid values are: "${authStrategyValues}"`
```
sansan, line 15: Maybe a bit more user friendly way would be to remind that they should include the auth strategy in strategies.consts
e.g. "Did you forget to add the auth strategy to ~/auth/strategies/strategies.consts"

src/auth/auth.service.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * @param apiKey - key to delete
+   */
+  public async deleteApiKey({ apiKey }: DeleteApiKeyDto): Promise<void> {
+    const user = this.getUserByApiKey(apiKey);
+    if (user) {
+      user.apiKeys = user.apiKeys.filter(key => key !== apiKey);
+    } else {
+      throw new NotFoundException('API key was not found');
```
sansan, line 49: imho we don't have to throw an error on delete if not found - we can just return 200 - as there is no record? 
e.g. if there are two parallel calls to delete the same API key - the response for both should be the same

src/middleware/auth.middleware.ts
```diff
@@ -0,0 +1,26 @@
...
+  use(req: unknown, res: unknown, next: () => void): void {
+    passport.authenticate(this.strategies, { session: false }, (request, user: User) => {
+      if (user) {
+        next();
+      } else {
+        throw new UnauthorizedException('REST API user was not found');
+      }
+    })(req, res, next);
+  }
+}
```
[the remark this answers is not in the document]
polymath-eric: My understanding is that guards are recommended for authorization, where as middleware is fine for authentication.

>Middleware is a fine choice for authentication, since things like token validation and attaching properties to the request object are not strongly connected with a particular route context (and its metadata).

https://docs.nestjs.com/guards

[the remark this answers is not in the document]
sansan: https://docs.nestjs.com/security/authentication#enable-authentication-globally :)

src/auth/dto/create-api-key.dto.ts
```diff
@@ -5,7 +5,7 @@ import { IsString } from 'class-validator';
 
 export class CreateApiKeyDto {
   @ApiProperty({
-    description: 'The user to create the api key for',
+    description: 'The id of the user to create the api key for',
```
prashantasdeveloper, line 8:
```suggestion
    description: 'The ID of the user to create the api key for',
```
nit

src/middleware/auth.middleware.ts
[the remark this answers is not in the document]
polymath-eric: That link depends on having this call:

```
return ctx.getContext().req.user;
```

Between line 19,20 I would need to add: 
```ts
request.user = user // or user.id
```

and then I'd be able to use that guard. I think I need middleware and guards.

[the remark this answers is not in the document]
polymath-eric: I think I was looking at the wrong header.

I will try with the guard and see how that goes.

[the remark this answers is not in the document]
polymath-eric: It took me a while to figure out, but this is now using the guard and is much less code compared to defining custom middleware.

src/auth/auth.controller.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  @ApiOperation({
+    summary: 'Create API Key',
+    description: 'This endpoint will create an API Key',
+  })
+  @ApiOkResponse({
+    description: 'Details of the API key created',
+    type: ApiKeyModel,
+  })
+  @Post('api-key/create')
```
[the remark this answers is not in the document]
polymath-eric: For now I am concerned with just finding the user.

I think with some authorization checks that could be something we can add in later. The open strategy gives a special user, so we could guard against that.

For now I'd say turn off open if its going to be exposed to the public.

polymath-eric:
@prashantasdeveloper does this look good to you? Is it OK to merge?

prashantasdeveloper (approved):

PR #122, opened 2022-10-17
feat: 🎸 Endpoint to get account subsidy

prashantasdeveloper:
### JIRA Link 

DA-6

### Changelog / Description 

Adds endpoint `/accounts/:account/subsidy` to fetch account subsidy

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/accounts/accounts.controller.ts
+++ b/src/accounts/accounts.controller.ts
@@ -1,5 +1,6 @@
-import { Body, Controller, Get, Param, Post, Query } from '@nestjs/common';
+import { Body, Controller, Get, HttpStatus, Param, Post, Query, Res } from '@nestjs/common';
 import {
+  ApiNoContentResponse,
   ApiNotFoundResponse,
   ApiOkResponse,
   ApiOperation,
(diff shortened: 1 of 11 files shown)


src/test-utils/mocks.ts
```diff
@@ -16,6 +18,13 @@ export type Mocked<T> = T &
       : T[K];
   };
 
+export const mockResponseObject = (): DeepMocked<Response> => {
```
polymath-eric, line 21:
```suggestion
export const createMockResponseObject = (): DeepMocked<Response> => {
```

src/accounts/accounts.controller.ts
```diff
@@ -122,4 +125,36 @@ export class AccountsController {
...
+  @ApiOkResponse({
+    description: 'Subsidy details for the Account',
+    type: SubsidyModel,
+  })
+  @ApiNoContentResponse({
+    description: 'Account is not being subsidized',
+  })
+  @ApiNotFoundResponse({
+    description: 'Account is not associated with any Identity',
+  })
```
polymath-eric, line 149: I think this *should* throw BadRequest for bad SS58 format, but its returning a InternalServerError in that case.

I think that error handling should be moved into the shared function instead of accounts service

polymath-eric, line 149:
```suggestion
```
I get the no content when I try, which I think is the right behavior.

polymath-eric (changes requested):

src/accounts/accounts.controller.ts
prashantasdeveloper: This was already handled in `findOne` method. I tested it out. Apparently await was missing which was causing this

polymath-eric (approved):

PR #123, opened 2022-10-18
Feat/da 63

sansan:
### JIRA Link 

https://polymesh.atlassian.net/browse/DA-63

### Changelog / Description 

- renames ComplianceController to ComplianceRequirementsController (incl. services and module)
- adds pause and unpause endpoints for compliance requirements

### Checklist - 

- [x] New Feature ?


src/compliance-requirements/compliance-requirements.controller.ts
```diff
@@ -72,4 +73,50 @@ export class ComplianceRequirementsController {
...
+  @ApiOperation({
+    summary: 'Pause Compliance requirements for an Asset',
+    description: 'This endpoint pauses Compliance rules for an Asset.',
+  })
+  @ApiParam({
+    name: 'ticker',
+    description: 'The ticker of the Asset whose compliance requirements are to be paused',
+    type: 'string',
+    example: 'TICKER',
+  })
```
polymath-eric, line 86: This should get `@ApiNotFoundResponse` in the case the asset isn't found.

polymath-eric (changes requested):
Overall looks good to me.

I think commit  `c33a7c1b908e107fe605df0ea8ed57272f0c44ba` should be only have renamed the controller.

I think Prashant is suggesting multiple controllers in one module so the routes can be grouped easier.

src/compliance-requirements/compliance-requirements.controller.ts
polymath-eric: There is also a permissions error if the signer doesn't own the asset. I am not sure what that will throw.

Let me know if you need help setting up the REST API to use different signers to try that out. (Make the asset with one signer, and try to pause it with another)

src/compliance-requirements/compliance-requirements.controller.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      const result = await controller.pauseRequirements(
+        { ticker },
+        validBody as TransactionBaseDto
+      );
+      expect(result).toEqual(response);
+    });
+  });
+
+  describe('unpauseRequirements', () => {
+    it('should accept SetRulesDto and set new Asset Compliance Rules', async () => {
```
prashantasdeveloper, line 90: Description got copied over

src/compliance-requirements/compliance-requirements.controller.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        transactions: [],
+      };
+      mockService.setRequirements.mockResolvedValue(response);
+      const result = await controller.setRequirements({ ticker }, validBody as SetRequirementsDto);
+      expect(result).toEqual(response);
+    });
+  });
+
+  describe('pauseRequirements', () => {
+    it('should accept SetRulesDto and set new Asset Compliance Rules', async () => {
```
prashantasdeveloper, line 76: Description got copied over

src/compliance-requirements/compliance-requirements.controller.ts
```diff
@@ -72,4 +73,50 @@ export class ComplianceRequirementsController {
     const result = await this.complianceRequirementsService.setRequirements(ticker, params);
     return handleServiceResult(result);
   }
+
+  @ApiOperation({
+    summary: 'Pause Compliance requirements for an Asset',
```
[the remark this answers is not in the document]
sansan: Changed, though - then we should change it also in other places as well as on other methods it's written in capital casing :) 

can do that :)

src/compliance-requirements/compliance-requirements.controller.ts
sansan: The tx service can throw an error (though haven't yet wrapped my head around how it will happen for subscribe on ws) - so added a swagger decorator to combine the possible error responses

src/compliance-requirements/compliance-requirements.controller.spec.ts
sansan: fixed

src/compliance-requirements/compliance-requirements.controller.spec.ts
sansan: fixed

src/compliance-requirements/compliance-requirements.controller.ts
polymath-eric: > though haven't yet wrapped my head around how it will happen for subscribe on ws

I am assuming you mean the subscription part as in when `webhookUrl` is passed and `ws` is for `websocket`? 

In that case we still run the sdk transaction validation layer (e.g. 'prepareAsset') and any errors like that will be handled as a non subscribed version. (hopefully this catches most issues, since the user will still pay gas fees if the SDK missed something)

If the SDK missed a chain constraint in the validation step (or the state changes between check and submission) the "async" error will need to be posted to the webhook subscriber, although we'd have returned 2xx to the submitter. 

In this service the websocket is an implementation detail of communicating with a  chain node, where as the SubscriptionService here uses regular HTTP `POST` for the public API.

src/common/decorators/swagger.ts
```diff
@@ -111,3 +115,26 @@ export function ApiTransactionResponse(
...
+  return applyDecorators(
+    ApiNotFoundResponse(options),
+    ApiBadRequestResponse({
+      description: 'Returned if transaction validation failed or same data already exists in chain',
+    }),
+    ApiUnprocessableEntityResponse({
+      description:
+        'Returned if there is insufficient balance to perform action or required perquisites for transaction have not been met',
+    }),
+    ApiInternalServerErrorResponse({
```
polymath-eric, line 136: I don't think we should document anything in 500 range, since they are *supposed* to be unexpected.

src/compliance/compliance-requirements.controller.ts
```diff
@@ -32,12 +39,15 @@ export class ComplianceController {
       'List of Compliance Requirements of the Asset along with Default Trusted Claim Issuers',
     type: ComplianceRequirementsModel,
   })
+  @ApiNotFoundResponse({
+    description: 'The asset was not found',
```
polymath-eric, line 43: This should be "The Asset was not found". 

Asset is an entity so it gets capitalized.

polymath-eric (commented):
looks good, two nitty comments

src/compliance/compliance-requirements.controller.ts
polymath-eric: there are a few places to be updated

prashantasdeveloper:
What do you do when you push new changes? Apparently commits are being pushed again and again with _fixup_ prefixed in commit names

polymath-eric:
The part before the emoji shouldn't be changed.

Usually I use chore, style or docs for subsequent commits. They should be squashed, but if they do get merged in the history should still make sense. For example if you fixed a bug only on your branch, that bug fix shouldn't show up as release notes.

My preference would be too rebase, but that should never be done after opening a PR on Github since it can make comments get lost. 

If you are force pushing please use `--force-with-lease` to not accidentally overwrite someone's else's commits. You shouldn't need to force push that often with our workflow though.

@sansan

src/compliance/compliance-requirements.controller.ts
polymath-eric, line 43:
```suggestion
    description: 'The Asset was not found',
```

src/compliance/compliance-requirements.controller.ts
```diff
@@ -64,12 +74,67 @@ export class ComplianceController {
     description: 'Details of the transaction',
     type: TransactionQueueModel,
   })
+  @ApiNotFoundResponse({
+    description: 'The asset was not found',
```
polymath-eric, line 78:
```suggestion
    description: 'The Asset was not found',
```

polymath-eric (approved):
LGTM

PR #128, opened 2022-10-27
 feat: 🎸 NetworkModule - getLatestBlock, getNetworkProperties

sansan:
### JIRA Link 

[DA-109
](https://polymesh.atlassian.net/browse/DA-109)
[DA-107](https://polymesh.atlassian.net/browse/DA-107)

### Changelog / Description 

- Adds NetworkModule
- Adds the ability to fetch the NetworkProperties
- Adds the ability to fetch latest block ID 

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/network/network.controller.spec.ts
+++ b/src/network/network.controller.spec.ts
@@ -41,7 +41,7 @@ describe('NetworkController', () => {
   });
 
   describe('getLatestBlock', () => {
-    it('should return latest block ID as NetworkBlockModel', async () => {
+    it('should return latest block number as NetworkBlockModel', async () => {
       const mockLatestBlock = new BigNumber(1);
       const mockResult = new NetworkBlockModel({ id: mockLatestBlock });
       mockNetworkService.getLatestBlock.mockResolvedValue(mockLatestBlock);
--- a/src/network/network.controller.ts
+++ b/src/network/network.controller.ts
@@ -26,11 +26,11 @@ export class NetworkController {
   }
 
   @ApiOperation({
-    summary: 'Get the latest block number',
-    description: 'This endpoint will provide the latest block ID',
+    summary: 'Get the latest block',
+    description: 'This endpoint will provide the latest block number',
   })
   @ApiOkResponse({
-    description: 'Latest block ID that has been added to the chain',
+    description: 'Latest block number that has been added to the chain',
     type: NetworkBlockModel,
   })
   @Get('latest-block')


src/network/network.controller.ts
```diff
@@ -0,0 +1,39 @@
+import { Controller, Get } from '@nestjs/common';
+import { ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
+
+import { NetworkBlockModel } from '~/network/models/network-block.model';
+import { NetworkPropertiesModel } from '~/network/models/network-properties.model';
+import { NetworkService } from '~/network/network.service';
+@ApiTags('network')
```
polymath-eric, line 7:
```suggestion
import { NetworkService } from '~/network/network.service';

@ApiTags('network')
```

src/network/network.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+@Controller('network')
+export class NetworkController {
+  constructor(private readonly networkService: NetworkService) {}
+
+  @ApiOperation({
+    summary: 'Fetch network properties',
+    description: 'This endpoint will provide network name and version',
```
polymath-eric, line 14:
```suggestion
    description: 'This endpoint will provide the network name and version',
```

src/network/network.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+  })
+  @ApiOkResponse({
+    description: 'Network properties response',
+    type: NetworkPropertiesModel,
+  })
+  @Get()
+  public async getNetworkProperties(): Promise<NetworkPropertiesModel> {
+    return new NetworkPropertiesModel(await this.networkService.getNetworkProperties());
```
polymath-eric, line 22:
```suggestion
    const networkProperties = await this.networkService.getNetworkProperties()
    return new NetworkPropertiesModel(networkProperties);
```

As a style thing, we never inline await

src/network/network.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @ApiOperation({
+    summary: 'Get the latest block number',
+    description: 'This endpoint will provide the latest block ID',
+  })
+  @ApiOkResponse({
+    description: 'Latest block ID that has been added to the blockchain',
```
polymath-eric, line 30:
```suggestion
    description: 'Latest block ID that has been added to the chain',
```

(nit)

src/network/network.service.spec.ts
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable import/first */
```
polymath-eric, line 1:
```suggestion
```

I don't think you need this here

src/network/network.module.ts
```diff
@@ -0,0 +1,15 @@
+/* istanbul ignore file */
+
+import { Module } from '@nestjs/common';
+
+import { NetworkController } from '~/network/network.controller';
+import { NetworkService } from '~/network/network.service';
+import { PolymeshModule } from '~/polymesh/polymesh.module';
+
+@Module({
```
polymath-eric, line 9: I know not many modules have them, but can you add a short TSDoc comment here?
```ts
/**
* responsible for the exposing the SDK network namespace
*/
```
Then it gives a nicer hover over card when its being imported into App module instead of `import NetworkModule`

src/network/network.service.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    networkService = module.get<NetworkService>(NetworkService);
+    polymeshService = module.get<PolymeshService>(PolymeshService);
+  });
+
+  afterEach(async () => {
+    await polymeshService.close();
+  });
+
+  it('should be defined', () => {
+    console.log(networkService);
```
polymath-eric, line 37:
```suggestion
```

We should probably add a lint rule for console.log. I've committed a few myself.

polymath-eric (commented):
Looks good overall. Mostly small nitty comments

polymath-eric (approved):

src/network/models/network-block.model.ts
```diff
@@ -0,0 +1,19 @@
+/* istanbul ignore file */
+import { ApiProperty } from '@nestjs/swagger';
+import { BigNumber } from '@polymeshassociation/polymesh-sdk';
+
+import { FromBigNumber } from '~/common/decorators/transformation';
+
+export class NetworkBlockModel {
```
prashantasdeveloper, line 7: IMO we don't require this model. We can directly send in the block number. WDUT @polymath-eric ?

src/network/network.controller.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+      mockNetworkService.getNetworkProperties.mockResolvedValue(mockResult);
+
+      const result = await controller.getNetworkProperties();
+
+      expect(result).toEqual(mockResult);
+    });
+  });
+
+  describe('getLatestBlock', () => {
+    it('should return latest block ID as NetworkBlockModel', async () => {
```
prashantasdeveloper, line 44:
```suggestion
    it('should return latest block number as NetworkBlockModel', async () => {
```

src/network/network.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+  public async getNetworkProperties(): Promise<NetworkPropertiesModel> {
+    const networkProperties = await this.networkService.getNetworkProperties();
+
+    return new NetworkPropertiesModel(networkProperties);
+  }
+
+  @ApiOperation({
+    summary: 'Get the latest block number',
+    description: 'This endpoint will provide the latest block ID',
+  })
```
prashantasdeveloper, line 31:
```suggestion
  @ApiOperation({
    summary: 'Get the latest block',
    description: 'This endpoint will provide the latest block number',
  })
```

src/network/network.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+  @ApiOkResponse({
+    description: 'Latest block ID that has been added to the chain',
```
prashantasdeveloper, line 33:
```suggestion
    description: 'Latest block number that has been added to the chain',
```

prashantasdeveloper (changes requested):
Left some nit comments around ID vs number changes

src/network/models/network-block.model.ts
polymath-eric: I think other things could go into this model though. We could return `23454` and technically its valid JSON, but generally people expect an object to be returned.

My preference would be to leave this. It can get more stuff added like block hash, number of events etc. and will help make us consistent

polymath-eric: The SDK method should have probably returned the full block for this, not just the number.

We can still do a second query if we want to support a more full model so I don't think its an issue, although not optimal.

polymath-eric: Object return value is much easier to extend. A raw primitive value is impossible to change without it being a major version change or new endpoint

prashantasdeveloper (approved):

PR #135, opened 2022-11-04
 feat: 🎸 trusted claim issuers: set, add, remove

sansan:
### JIRA Link 
https://polymesh.atlassian.net/browse/DA-61

### Changelog / Description 

- moves list trusted claim issuers to a separate controller
- Adds endpoints to add, set and remove trusted claim issuers

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/compliance/dto/trusted-claim-issuer.dto.ts
+++ b/src/compliance/dto/trusted-claim-issuer.dto.ts
@@ -11,10 +11,11 @@ export class TrustedClaimIssuerDto {
     enum: ClaimType,
     isArray: true,
     nullable: true,
+    default: null,
   })
   @IsOptional()
   @IsEnum(ClaimType, { each: true })
-  readonly trustedFor?: ClaimType[] | null;
+  readonly trustedFor: ClaimType[] | null;
 
   @ApiPropertyOptional({
     description: 'The Identity of the Claim Issuer',
--- a/src/compliance/trusted-claim-issuers.service.ts
+++ b/src/compliance/trusted-claim-issuers.service.ts
@@ -1,10 +1,5 @@
 import { Injectable } from '@nestjs/common';
-import {
-  Asset,
-  ModifyAssetTrustedClaimIssuersAddSetParams,
-  ModifyAssetTrustedClaimIssuersRemoveParams,
-  TrustedClaimIssuer,
-} from '@polymeshassociation/polymesh-sdk/types';
+import { Asset, TrustedClaimIssuer } from '@polymeshassociation/polymesh-sdk/types';
 
 import { AssetsService } from '~/assets/assets.service';
 import { ServiceReturn } from '~/common/utils';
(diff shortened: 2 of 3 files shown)


src/common/decorators/swagger.ts
```diff
@@ -126,11 +126,15 @@ type SupportedHttpStatusCodes =
  * @param messages - key value map of HTTP response code to their description that will be passed to appropriate `MethodDecorator`
  */
 export function ApiTransactionFailedResponse(
-  messages: Partial<Record<SupportedHttpStatusCodes, string>>
+  messages: Partial<Record<SupportedHttpStatusCodes, string | string[]>>
```
polymath-eric, line 129:
```suggestion
  messages: Partial<Record<SupportedHttpStatusCodes, string[]>>
```
It's not a big deal adding in an extra `[]` for a single item and then it's easier to add a second.

src/test-utils/service-mocks.ts
```diff
@@ -180,6 +179,13 @@ export class MockAuthService {
   validateApiKey = jest.fn();
 }
 
+export class MockTrustedClaimIssuersService {
```
polymath-eric, line 182: Prashant added this package recently: https://www.npmjs.com/package/@golevelup/ts-jest

You should be able to do `createMock<MockTrustedClaimIssuersService>()` where ever you need this instead.

You might want to define a Nest `Provder` object in this file to save on test boiler plate.

polymath-eric (commented):
Can you update this to be against `feature-release` (or `master` if it gets merged)?

src/common/decorators/swagger.ts
polymath-eric: `Arrary.isArrary` -> `rawDescription.length > 1` ?

I won't block approval on this though, either way is OK

src/common/decorators/swagger.ts
```diff
@@ -126,11 +126,15 @@ type SupportedHttpStatusCodes =
...
 ): ReturnType<typeof applyDecorators> {
   const decorators: MethodDecorator[] = [];
 
-  Object.entries(messages).forEach(([statusCode, description]) => {
+  Object.entries(messages).forEach(([statusCode, rawDescription]) => {
+    const description = Array.isArray(rawDescription)
+      ? '<ul>' + rawDescription.map(v => `<li>${v}</li>`).join(' ') + '</ul>'
```
polymath-eric, line 135: (style) Usually we don't do single letter variables, even if they are close to the definition.
```suggestion
      ? '<ul>' + rawDescription.map(description => `<li>${description}</li>`).join(' ') + '</ul>'
```

I still find myself doing single letters for things like this from time to time though.

prashantasdeveloper (changes requested):
While your at it, the `trustedFor` in `TrustedClaimIssuerDto` needs to be amended to support `null`. Can you take care of that as well?

src/compliance/trusted-claim-issuers.controller.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+        claimIssuers: [],
+        signer: 'Alice',
+      };
+
+      const response = {
+        transactions: [transaction],
+      };
+
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
polymath-eric, line 125: why ts ignore?

src/compliance/dto/trusted-claim-issuer.dto.ts
```diff
@@ -10,10 +10,11 @@ export class TrustedClaimIssuerDto {
       'List of Claim types for which an Identity is trusted for verifying. Defaults to all types',
     enum: ClaimType,
     isArray: true,
+    nullable: true,
   })
   @IsOptional()
   @IsEnum(ClaimType, { each: true })
-  readonly trustedFor?: ClaimType[];
+  readonly trustedFor?: ClaimType[] | null;
```
polymath-eric, line 17:
```suggestion
  readonly trustedFor: ClaimType[] | null = null;
```
If this gets defaulted to null, then we don't have to deal with the undefined case, which means a type assertion isn't needed in the service.

src/compliance/dto/trusted-claim-issuer.dto.ts
```diff
@@ -10,10 +10,11 @@ export class TrustedClaimIssuerDto {
       'List of Claim types for which an Identity is trusted for verifying. Defaults to all types',
     enum: ClaimType,
     isArray: true,
+    nullable: true,
```
polymath-eric, line 13:
```suggestion
    nullable: true,
    default: null
```

src/compliance/trusted-claim-issuers.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    return asset.compliance.trustedClaimIssuers.get();
+  }
+
+  public async set(ticker: string, params: SetTrustedClaimIssuersDto): ServiceReturn<Asset> {
+    const { signer, webhookUrl, ...rest } = params;
+    const asset = await this.assetsService.findOne(ticker);
+
+    return this.transactionsService.submit(
+      asset.compliance.trustedClaimIssuers.set,
+      rest as ModifyAssetTrustedClaimIssuersAddSetParams,
```
polymath-eric, line 34:
```suggestion
      rest,
```
If the DTO matches the signature there is no need to assert this.

The SDK is checking explicitly for null in a function, so this assertion is unsound.

`(aTrustedFor === null && bTrustedFor === null)`
https://github.com/PolymathNetwork/polymesh-sdk/blob/339b750301f99ccff0c390a00fdbf1f153765999/src/api/procedures/modifyAssetTrustedClaimIssuers.ts#L84

src/test-utils/mocks.ts
```diff
@@ -6,18 +6,33 @@ import {
...
       : T[K];
   };
 
+export const mockTrustedClaimIssuer = createMock<TrustedClaimIssuer<true>>();
+
+export const createMockTransactionResult = <T>({
+  transactions,
+  result,
+}: {
+  transactions: TransactionResult<any>['transactions'];
```
polymath-eric, line 30:
```suggestion
  transactions: TransactionResult<T>['transactions'];
```
?

polymath-eric (approved):
LGTM

prashantasdeveloper (approved):

PR #145, opened 2022-11-23
feat: 🎸 Add Asset Metadata endpoints

prashantasdeveloper:
### JIRA Link 

DA-446, DA-447, DA-448, DA-449, DA-451, DA-452

### Changelog / Description 

Add Asset Metadata endpoints

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/metadata/dto/create-metadata.dto.ts
+++ b/src/metadata/dto/create-metadata.dto.ts
@@ -27,7 +27,7 @@ export class CreateMetadataDto extends TransactionBaseDto {
   @ApiPropertyOptional({
     description: 'Value for the Asset Metadata',
     type: 'string',
-    example: 'SOME_VALUE',
+    example: 'Some value',
   })
   @ValidateIf(({ details, value }: CreateMetadataDto) => !!value || !!details)
   @IsString()
--- a/src/metadata/dto/create-metadata.dto.ts
+++ b/src/metadata/dto/create-metadata.dto.ts
@@ -1,0 +1,44 @@
+/* istanbul ignore file */
+
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { Type } from 'class-transformer';
+import { IsOptional, IsString, ValidateIf, ValidateNested } from 'class-validator';
+
+import { TransactionBaseDto } from '~/common/dto/transaction-base-dto';
+import { MetadataSpecDto } from '~/metadata/dto/metadata-spec.dto';
+import { MetadataValueDetailsDto } from '~/metadata/dto/metadata-value-details.dto';
+
+export class CreateMetadataDto extends TransactionBaseDto {
+  @ApiProperty({
+    description: 'Name of the Asset Metadata',
+    example: 'Maturity',
+  })
+  @IsString()
+  readonly name: string;
+
+  @ApiProperty({
+    description: 'Details about the Asset Metadata',
+    type: MetadataSpecDto,
+  })
+  @ValidateNested()
+  @Type(() => MetadataSpecDto)
+  readonly specs: MetadataSpecDto;
+
+  @ApiPropertyOptional({
+    description: 'Value for the Asset Metadata',
+    type: 'string',
+    example: 'SOME_VALUE',
+  })
+  @ValidateIf(({ details, value }: CreateMetadataDto) => !!value || !!details)
+  @IsString()
+  readonly value?: string;
+
+  @ApiPropertyOptional({
+    description: 'Details about the Asset Metadata value',
+    type: MetadataValueDetailsDto,
+  })
+  @IsOptional()
+  @ValidateNested()
+  @Type(() => MetadataValueDetailsDto)
+  readonly details?: MetadataValueDetailsDto;
+}
--- a/src/metadata/dto/metadata-spec.dto.ts
+++ b/src/metadata/dto/metadata-spec.dto.ts
@@ -7,14 +7,14 @@ export class MetadataSpecDto {
   @ApiPropertyOptional({
     description: 'Off-chain specs or documentation link',
     type: 'string',
-    example: 'https://www.someexample.com',
+    example: 'https://example.com',
   })
   @IsOptional()
   @IsString()
   readonly url?: string;
 
   @ApiPropertyOptional({
-    description: 'Description of metadata type',
+    description: 'Description of the Metadata type',
     type: 'string',
     example: 'Some description',
   })
@@ -23,7 +23,8 @@ export class MetadataSpecDto {
   readonly description?: string;
 
   @ApiPropertyOptional({
-    description: 'SCALE encoded `AssetMetadataTypeDef`',
+    description:
+      '[SCALE](https://wiki.polkadot.network/docs/build-tools-index#scale-codec) encoded `AssetMetadataTypeDef`',
     type: 'string',
     example: 'Some example',
   })
--- a/src/metadata/dto/metadata-spec.dto.ts
+++ b/src/metadata/dto/metadata-spec.dto.ts
@@ -1,0 +1,33 @@
+/* istanbul ignore file */
+
+import { ApiPropertyOptional } from '@nestjs/swagger';
+import { IsOptional, IsString } from 'class-validator';
+
+export class MetadataSpecDto {
+  @ApiPropertyOptional({
+    description: 'URL describing the Asset metadata',
+    type: 'string',
+    example: 'https://www.someexample.com',
+  })
+  @IsOptional()
+  @IsString()
+  readonly url?: string;
+
+  @ApiPropertyOptional({
+    description: 'Description about the Asset metadata',
+    type: 'string',
+    example: 'https://www.someexample.com',
+  })
+  @IsOptional()
+  @IsString()
+  readonly description?: string;
+
+  @ApiPropertyOptional({
+    description: 'Type definition describing the metadata',
+    type: 'string',
+    example: 'https://www.someexample.com',
+  })
+  @IsOptional()
+  @IsString()
+  readonly typedef?: string;
+}
--- a/src/metadata/dto/metadata-spec.dto.ts
+++ b/src/metadata/dto/metadata-spec.dto.ts
@@ -5,7 +5,7 @@ import { IsOptional, IsString } from 'class-validator';
 
 export class MetadataSpecDto {
   @ApiPropertyOptional({
-    description: 'URL describing the Asset metadata',
+    description: 'Off-chain specs or documentation link',
     type: 'string',
     example: 'https://www.someexample.com',
   })
@@ -14,18 +14,18 @@ export class MetadataSpecDto {
   readonly url?: string;
 
   @ApiPropertyOptional({
-    description: 'Description about the Asset metadata',
+    description: 'Description of metadata type',
     type: 'string',
-    example: 'https://www.someexample.com',
+    example: 'Some description',
   })
   @IsOptional()
   @IsString()
   readonly description?: string;
 
   @ApiPropertyOptional({
-    description: 'Type definition describing the metadata',
+    description: 'SCALE encoded `AssetMetadataTypeDef`',
     type: 'string',
-    example: 'https://www.someexample.com',
+    example: 'Some example',
   })
   @IsOptional()
   @IsString()
--- a/src/metadata/metadata.controller.spec.ts
+++ b/src/metadata/metadata.controller.spec.ts
@@ -18,7 +18,7 @@ import { CreatedMetadataEntryModel } from '~/metadata/models/created-metadata-en
 import { MetadataDetailsModel } from '~/metadata/models/metadata-details.model';
 import { MetadataEntryModel } from '~/metadata/models/metadata-entry.model';
 import { MetadataValueModel } from '~/metadata/models/metadata-value.model';
-import { createMockTransactionResult, MockMetadataEntry } from '~/test-utils/mocks';
+import { createMockMetadataEntry, createMockTransactionResult } from '~/test-utils/mocks';
 import { mockMetadataServiceProvider } from '~/test-utils/service-mocks';
 
 describe('MetadataController', () => {
@@ -49,11 +49,8 @@ describe('MetadataController', () => {
 
   describe('getMetadata', () => {
     it('should return the list of all metadata for a given ticker', async () => {
-      const mockMetadataEntry = new MockMetadataEntry();
-      when(mockService.findAll)
-        .calledWith(ticker)
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        .mockResolvedValue([mockMetadataEntry as any]);
+      const mockMetadataEntry = createMockMetadataEntry();
+      when(mockService.findAll).calledWith(ticker).mockResolvedValue([mockMetadataEntry]);
 
       const result = await controller.getMetadata({ ticker });
 
@@ -65,7 +62,7 @@ describe('MetadataController', () => {
 
   describe('getSingleMetadata', () => {
     it('should return the Metadata details for a specific type and ID', async () => {
-      const mockMetadataEntry = new MockMetadataEntry();
+      const mockMetadataEntry = createMockMetadataEntry();
       const mockDetails = {
         name: 'Some metadata',
         specs: {
@@ -85,8 +82,7 @@ describe('MetadataController', () => {
 
       when(mockService.findOne)
         .calledWith({ ticker, type, id })
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        .mockResolvedValue(mockMetadataEntry as any);
+        .mockResolvedValue(mockMetadataEntry);
 
       const result = await controller.getSingleMetadata({ ticker, type, id });
 
@@ -111,8 +107,8 @@ describe('MetadataController', () => {
         type: TransactionType.Single,
         transactionTag: TxTags.asset.RegisterAssetMetadataLocalType,
       };
-      const mockMetadataEntry = new MockMetadataEntry();
-      const testTxResult = createMockTransactionResult<MockMetadataEntry>({
+      const mockMetadataEntry = createMockMetadataEntry();
+      const testTxResult = createMockTransactionResult<MetadataEntry>({
         transactions: [transaction],
         result: mockMetadataEntry,
       });
@@ -124,10 +120,7 @@ describe('MetadataController', () => {
         signer: 'Alice',
       };
 
-      when(mockService.create)
-        .calledWith(ticker, mockPayload)
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        .mockResolvedValue(testTxResult as any);
+      when(mockService.create).calledWith(ticker, mockPayload).mockResolvedValue(testTxResult);
 
       const result = await controller.createMetadata({ ticker }, mockPayload);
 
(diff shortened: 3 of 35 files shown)


src/metadata/metadata.controller.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+      expect(result).toEqual({
+        results: [new MetadataEntryModel({ asset: ticker, type, id })],
+      });
+    });
+  });
+
+  describe('getSingleMetadata', () => {
+    it('should return the Metadata details for a specific type and ID', async () => {
+      const mockMetadataEntry = new MockMetadataEntry();
```
polymath-eric, line 68: - use `createMock`, specify the assertion values
- remove `as any` cast

`mockResolvedValue` should be able to stay as they are

src/metadata/metadata.controller.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      );
+    });
+  });
+
+  describe('createMetadata', () => {
+    it('should accept CreateMetadataDto and create a local Asset Metadata for the given ticker', async () => {
+      const transaction = {
+        blockHash: '0x1',
+        transactionHash: '0x2',
+        blockNumber: new BigNumber(1),
```
polymath-eric, line 110: (optional) Just an idea, but we could put these into `src/test-utils/consts.ts` and call it something like `txBase` and do something like this:

```suggestion
        ...txBase
```

OK to leave as it though

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+      'This endpoint creates a local metadata for the given Asset. The metadata value can be set by passing `value` parameter and specifying other optional `details` about the value',
+  })
+  @ApiParam({
+    name: 'ticker',
+    description: 'The ticker of the Asset for which the metadata is to be created',
+    type: 'string',
+    example: 'TICKER',
+  })
+  @ApiTransactionResponse({
+    description: 'Newly created Metadata entry along with transaction details',
```
polymath-eric, line 103:
```suggestion
    description: 'The newly created Metadata entry along with transaction details',
```

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+    type: CreatedMetadataEntryModel,
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.NOT_FOUND]: ['The Asset was not found'],
+    [HttpStatus.BAD_REQUEST]: [
+      'Asset Metadata name length exceeded',
+      'Asset Metadata value length exceeded',
+    ],
+    [HttpStatus.UNPROCESSABLE_ENTITY]: [
+      'Asset Metadata with the given name already exists',
```
polymath-eric, line 113: Name exists should be `409 Conflict` imo

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+      new CreatedMetadataEntryModel({
+        transactions,
+        metadata: new MetadataEntryModel({ asset: assetTicker, id, type }),
+      });
+
+    return handleServiceResult(serviceResult, resolver);
+  }
+
+  @ApiOperation({
+    summary: 'Set Asset Metadata value and/or its details',
```
polymath-eric, line 142:
```suggestion
    summary: 'Set an Asset's Metadata value and details',
```

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+    description:
+      'This endpoint assigns a new value for the Metadata along with its details or optionally only set the details (expiry + lock status) of the Metadata value. Note that the value of a locked Metadata cannot be altered',
```
polymath-eric, line 144:
```suggestion
      'This endpoint assigns a new value for the Metadata along with its expiry and lock status (when provided with `details`) of the Metadata value. Note that the value of a locked Metadata cannot be altered',
```

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+  })
+  @ApiParam({
+    name: 'ticker',
+    description: 'The ticker of the Asset for which the metadata value is to be set',
```
polymath-eric, line 148:
```suggestion
    description: 'The ticker of the Asset for which the Metadata value is to be set',
```

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+    description: 'Details about the transaction',
+    type: TransactionQueueModel,
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.NOT_FOUND]: ['The Asset was not found'],
+    [HttpStatus.BAD_REQUEST]: ['Asset Metadata name length exceeded'],
+    [HttpStatus.UNPROCESSABLE_ENTITY]: [
+      'Details cannot be set for a locked Metadata value',
+      'Metadata value is currently locked',
+      'Metadata value details cannot be set for a metadata with no value',
```
polymath-eric, line 174:
```suggestion
      'Details cannot be set for a metadata without a value',
```
I think the sentence structure should be consistent with the locked error.

src/metadata/metadata.controller.ts
```diff
@@ -0,0 +1,186 @@
...
+    description: 'The ID of Asset Metadata',
+    type: 'string',
+    example: '1',
+  })
+  @ApiTransactionResponse({
+    description: 'Details about the transaction',
+    type: TransactionQueueModel,
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.NOT_FOUND]: ['The Asset was not found'],
```
polymath-eric, line 169: Could the Metadata not be found as well?

src/metadata/metadata.service.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+        let error;
+        try {
+          await service.findOne({ ticker, type, id });
+        } catch (err) {
+          error = err;
+        }
+
+        expect(error).toBeInstanceOf(NotFoundException);
+      });
+    });
```
polymath-eric, line 147:
```suggestion
    });
    
```

src/metadata/metadata.service.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+        mockAsset.metadata.getOne.mockImplementation(() => {
+          throw expectedError;
+        });
+
+        let error;
+        try {
+          await service.findOne({ ticker, type, id });
+        } catch (err) {
+          error = err;
+        }
```
polymath-eric, line 160:
```suggestion
  await expect(service.findOne({ticker, type, id}).rejects.toThrow(expectedError)
```

src/metadata/metadata.service.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+
+        expect(error).toEqual(expectedError);
+
+        expectedError = new Error('Something else');
+
+        mockIsPolymeshError.mockReturnValue(true);
+
+        error = null;
+        try {
```
polymath-eric, line 169: `expect().rejects` syntax here to

src/metadata/metadata.service.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+          await service.findOne({ ticker, type, id });
+        } catch (err) {
+          error = err;
+        }
+
+        expect(error).toEqual(expectedError);
+      });
+    });
+    describe('otherwise', () => {
+      it('should return the Asset', async () => {
```
polymath-eric, line 179:
```suggestion
      it('should return the Metadata entry', async () => {
```

src/metadata/metadata.service.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+
+      const mockAsset = new MockAsset();
+      when(mockAssetsService.findOne).calledWith(ticker).mockResolvedValue(mockAsset);
+
+      mockTransactionsService.submit.mockResolvedValue({
+        result: mockMetadataEntry,
+        transactions: [mockTransaction],
+      });
+
+      const signer = 'signer';
```
polymath-eric, line 209:
```suggestion
```
import `signer` from `src/test-utils/consts.ts` instead

src/metadata/metadata.service.spec.ts
```diff
@@ -0,0 +1,276 @@
...
+        .calledWith({ ticker, type, id })
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        .mockResolvedValue(mockMetadataEntry as any);
+
+      mockTransactionsService.submit.mockResolvedValue({
+        result: mockMetadataEntry,
+        transactions: [mockTransaction],
+      });
+
+      const signer = 'signer';
```
polymath-eric, line 253:
```suggestion
```
import from test consts

src/metadata/metadata.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    } catch (err: unknown) {
+      if (isPolymeshError(err)) {
+        const { code, message } = err;
+        if (code === ErrorCode.DataUnavailable) {
+          throw new NotFoundException(message);
+        }
+      }
+
+      throw err;
+    }
```
polymath-eric, line 50: This error logic is duplicated. This should work instead

```ts
    return await metadata.getOne({ type, id }).catch(handleSdkError);
```

src/metadata/metadata.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  public async setValue(
+    params: MetadataParamsDto,
+    body: SetMetadataDto
+  ): ServiceReturn<MetadataEntry> {
+    const { signer, webhookUrl, ...rest } = body;
+
+    const { set } = await this.findOne(params);
+
+    return this.transactionsService.submit(set, rest as SetMetadataParams, {
```
polymath-eric, line 74: The `as SetMetadataParams` makes me a bit nervous, but its not obvious to me how to get rid of it.

If it needs to stay than so be it.

src/metadata/models/metadata-entry.model.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { ApiProperty } from '@nestjs/swagger';
+import { BigNumber } from '@polymeshassociation/polymesh-sdk';
+import { MetadataType } from '@polymeshassociation/polymesh-sdk/types';
+
+import { FromBigNumber } from '~/common/decorators/transformation';
+
+export class MetadataEntryModel {
+  @ApiProperty({
+    description: 'Ticker of the Asset for which this is the metadata',
```
polymath-eric, line 11:
```suggestion
    description: 'The ticker of the Asset for which this is the Metadata for',
```

src/metadata/models/metadata-entry.model.ts
```diff
@@ -0,0 +1,35 @@
...
+    type: 'string',
+    example: 'TICKER',
+  })
+  readonly asset: string;
+
+  @ApiProperty({
+    description: 'Type of metadata represented by this instance',
```
polymath-eric, line 18:
```suggestion
    description: 'The type of metadata represented by this instance',
```

src/metadata/models/metadata-spec.model.ts
```diff
@@ -0,0 +1,30 @@
+/* istanbul ignore file */
+
+import { ApiPropertyOptional } from '@nestjs/swagger';
+
+export class MetadataSpecModel {
+  @ApiPropertyOptional({
+    description: 'Off-chain specs or documentation link',
+    type: 'string',
+    example: 'https://www.someexample.com',
```
polymath-eric, line 9:
```suggestion
    example: 'https://example.com',
```

src/metadata/models/metadata-spec.model.ts
```diff
@@ -0,0 +1,30 @@
...
+  })
+  readonly url?: string;
+
+  @ApiPropertyOptional({
+    description: 'Description of metadata type',
+    type: 'string',
+    example: 'https://www.someexample.com',
```
polymath-eric, line 16:
```suggestion
    example: 'https://example.com',
```

src/metadata/models/metadata-spec.model.ts
```diff
@@ -0,0 +1,30 @@
...
+export class MetadataSpecModel {
+  @ApiPropertyOptional({
+    description: 'Off-chain specs or documentation link',
+    type: 'string',
+    example: 'https://www.someexample.com',
+  })
+  readonly url?: string;
+
+  @ApiPropertyOptional({
+    description: 'Description of metadata type',
```
polymath-eric, line 14:
```suggestion
    description: 'Description of the Metadata type',
```

src/metadata/models/metadata-spec.model.ts
```diff
@@ -0,0 +1,30 @@
...
+    type: 'string',
+    example: 'https://www.someexample.com',
+  })
+  readonly description?: string;
+
+  @ApiPropertyOptional({
+    description: 'SCALE encoded `AssetMetadataTypeDef`',
+    type: 'string',
+    example: 'https://www.someexample.com',
```
polymath-eric, line 23:
```suggestion
    example: 'https://example.com',
```

src/metadata/models/metadata-value.model.ts
```diff
@@ -0,0 +1,40 @@
...
+  })
+  readonly lockStatus: MetadataLockStatus;
+
+  @ApiPropertyOptional({
+    description:
+      'Date till which the Metadata value will be locked. This only applies when `status` is `LockedUntil`',
+    type: 'string',
+    example: new Date('05/23/2021').toISOString(),
+  })
+  readonly lockedUntil?: Date;
```
polymath-eric, line 35: To me this seems a bit inconsistent with `expiry` which is nullable.

I am OK with it as long as you know the difference in behavior and this is what you want: https://dev.to/jsmates/why-the-undefined-variable-is-not-sent-in-a-json-api-response-18k3#:~:text=Undefined%20is%20not%20a%20valid,valid%20JSON%20and%20vice%20versa.&text=All%20other%20values%20will%20be,undefined%20won't%20be%20sent.

src/transactions/transactions.service.ts
```diff
@@ -87,7 +87,7 @@ export class TransactionsService {
         );
       }
     } catch (error) {
-      handleSdkError(error);
+      handleSdkError(error as Error);
```
polymath-eric, line 90:
```suggestion
      handleSdkError(error);
```

This shouldn't be needed and doesn't provide any additional type safety

src/transactions/transactions.util.ts
```diff
@@ -97,7 +97,7 @@ export async function processTransaction<
       transactions: [assembleTransactionResponse(procedure)],
     };
   } catch (err) /* istanbul ignore next: not worth the trouble */ {
-    handleSdkError(err);
+    handleSdkError(err as Error);
```
polymath-eric, line 100:
```suggestion
    handleSdkError(err);
```

src/transactions/transactions.service.ts
polymath-eric: Seems like something changed that makes us need it now with new packages.

OK to leave

src/metadata/dto/create-metadata.dto.ts
```diff
@@ -0,0 +1,44 @@
...
+  @ValidateNested()
+  @Type(() => MetadataSpecDto)
+  readonly specs: MetadataSpecDto;
+
+  @ApiPropertyOptional({
+    description: 'Value for the Asset Metadata',
+    type: 'string',
+    example: 'SOME_VALUE',
+  })
+  @ValidateIf(({ details, value }: CreateMetadataDto) => !!value || !!details)
```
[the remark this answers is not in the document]
prashantasdeveloper: My idea was it to make it err if details are provided without value. `details` field actually attributes to `value` part of metadata.

src/metadata/dto/metadata-spec.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+    example: 'Some description',
+  })
+  @IsOptional()
+  @IsString()
+  readonly description?: string;
+
+  @ApiPropertyOptional({
+    description: 'SCALE encoded `AssetMetadataTypeDef`',
+    type: 'string',
+    example: 'Some example',
```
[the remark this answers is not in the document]
prashantasdeveloper: I have added the link using markdown

src/metadata/metadata.controller.ts
prashantasdeveloper: I have changed the text, but to be clear what we can do with this API is that we can 
* only set value
* set value along with details
* only set details (this requires the value to be already set)

src/metadata/models/metadata-value.model.ts
prashantasdeveloper: Wrt to SDK, `null` signifies never in such scenarios (more of like eternity). LockedUntil is only present when status would say `LockedUntil`, hence the `?` part. `null` value here would mean that LockedUntil eternity which isn't the case. So for this scenario it fits correctly IMO

src/metadata/metadata.service.ts
prashantasdeveloper: This is required in this case. To get rid of it we just need to create a new object (of type SetMetadataParams) with all the keys & values within rest which seems redundant

src/metadata/metadata.controller.spec.ts
prashantasdeveloper: After Toms change, we can refactor couples of files to match this pattern. For now I'll leave like this

src/metadata/metadata.controller.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    expect(controller).toBeDefined();
+  });
+
+  describe('getMetadata', () => {
+    it('should return the list of all metadata for a given ticker', async () => {
+      const mockMetadataEntry = new MockMetadataEntry();
+      when(mockService.findAll)
+        .calledWith(ticker)
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        .mockResolvedValue([mockMetadataEntry as any]);
```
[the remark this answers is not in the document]
prashantasdeveloper: That's how we have been setting up all the test suites, rather it be SDK or REST API. `TICKER` serves as a default value for Asset.ticker and we have created the metadata for the same as well. So I don't think that should be a problem. Said that I'll change the code to use createMock instead as well

src/test-utils/service-mocks.ts
```diff
@@ -204,6 +205,11 @@ export const mockTrustedClaimIssuersServiceProvider: ValueProvider<TrustedClaimI
   useValue: createMock<TrustedClaimIssuersService>(),
 };
 
+export const mockMetadataServiceProvider: ValueProvider<MetadataService> = {
```
VictorVicente, line 208:
```suggestion
export const MockMetadataServiceProvider: ValueProvider<MetadataService> = {
```

polymath-eric: This is a plain JavaScript object, so camel case seems right to me

Pascal case should be used for types and classes

src/metadata/dto/create-metadata.dto.ts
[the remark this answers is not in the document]
polymath-eric: Makes sense. I was thinking details was more related to the key part of the metadata

src/metadata/metadata.controller.ts
polymath-eric: Fair enough

polymath-eric (approved):

PR #147, opened 2022-11-29
feat: 🎸 Add subsidy related APIs

prashantasdeveloper:
### JIRA Link 

DA-111, DA-112, DA-113, DA-114

### Changelog / Description 

Add subsidy related APIs

### Checklist - 

- [x] New Feature ?


prashantasdeveloper:
Tests are failing as it depends on PR https://github.com/PolymeshAssociation/polymesh-sdk/pull/897.

src/subsidy/subsidy.controller.ts
```diff
@@ -0,0 +1,115 @@
...
+  @ApiTransactionResponse({
+    description: 'Details about the transaction',
+    type: TransactionQueueModel,
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.BAD_REQUEST]: [
+      'The Beneficiary Account already has a pending invitation to add this account as a subsidizer with the same allowance',
+    ],
+  })
+  @Post('allowance/modify')
```
polymath-eric, line 91: I think this should be 3 endpoints instead of nesting method dispatching in the body. Its more boiler plate and documentation, but is more consistent being broken up with one route, one method.

The service method can stay the same and the operation can be hard coded in the controller function in the call

src/subsidy/subsidy.controller.ts
```diff
@@ -0,0 +1,115 @@
...
+  async subsidizeAccount(@Body() params: CreateSubsidyDto): Promise<TransactionResponseModel> {
+    const serviceResult = await this.subsidyService.subsidizeAccount(params);
+
+    return handleServiceResult(serviceResult, authorizationRequestResolver);
+  }
+
+  @ApiOperation({
+    summary: 'Modify allowance for a Subsidy relationship an account',
+    description:
+      'This endpoint allows to set/increase/decrease allowance of a Subsidy relationship. Note that only the subsidizer is allowed to modify the allowance',
```
polymath-eric, line 80: Its not obvious why increase or decrease should be used. We ourselves missed the fact set can cause a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) style race condition originally.

I think we should have this be `/set` and mention why increase/decrease is preferred.

src/subsidy/subsidy.service.ts
```diff
@@ -0,0 +1,105 @@
...
+  public async getAllowance(beneficiary: string, subsidizer: string): Promise<BigNumber> {
+    const subsidy = this.findOne(beneficiary, subsidizer);
+
+    try {
+      return await subsidy.getAllowance();
+    } catch (err: unknown) {
+      if (isPolymeshError(err)) {
+        const { code } = err;
+        if (code === ErrorCode.DataUnavailable) {
+          throw new NotFoundException('The Subsidy no longer exists');
```
polymath-eric, line 98: I think its better to handle all the SDK errors in one place with the `handleSdkError` function. We can update that method to accept `unknown` instead of `Error` to avoid needing the `as Error` cast for the caller.

The SDK message should be descriptive enough to return directly, and if its not we should update that message.

To me this looks messy and is a lot of lines to repeat. The SDK errors should be good enough to be handled in.a generic manner imo.

That said I know this pattern exists already and won't block the PR for this.

src/transactions/transactions.service.ts
```diff
@@ -87,7 +87,7 @@ export class TransactionsService {
         );
       }
     } catch (error) {
-      handleSdkError(error);
+      handleSdkError(error as Error);
```
polymath-eric, line 90: I think `handleSdkError` should update its signature to be 

```
handleSdkError(err: unknown): never
```

then at the bottom of the function to:

```
  if (err instanceof Error) {
    throw new InternalServerErrorException(err.message);
  }

  throw err; // maybe throw new InternalServerErrorException('An unexpected error occurred') instead
````

**Review by polymath-eric — CHANGES_REQUESTED** [review]
I think the `modify` should be split up into 3 endpoints so our public API is consistent.

The service can stay as it is since thats an internal thing. Even the DTO can remain as it is since the operation enum can be hard coded in each controller.

File: src/subsidy/subsidy.service.ts (hunk 3 above)
Comment (prashantasdeveloper) [reply src/subsidy/subsidy.service.ts:98]: I agree with you. I'll change it here and I've created a separate task - [DA-468](https://polymesh.atlassian.net/browse/DA-468) to replace all other instances and cleanup the test cases for the same.

File: src/transactions/transactions.util.spec.ts [hunk 5]
```diff
@@ -51,13 +51,21 @@ describe('processTransaction', () => {
...
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      await expect(processTransaction(mockVenue.modify as any, {}, {})).rejects.toBeInstanceOf(
-        InternalServerErrorException
-      );
+
+      await expect(result).rejects.toBeInstanceOf(InternalServerErrorException);
+
+      mockVenue.modify.mockImplementationOnce(() => {
+        // eslint-disable-next-line no-throw-literal
+        throw 'Some unexpected error';
```
polymath-eric, line 65:
```suggestion
        throw new Error('Some unexpected error');
```

polymath-eric (approved):
LGTM, left one nit comment

polymath-eric:
looks like the tests might need a type assertion as well

src/transactions/transactions.util.spec.ts
prashantasdeveloper: I was trying to test something that is not an instance of Error

PR #151, opened 2022-12-14
feat: addClaims, editClaims, revokeClaims

sansan:
### JIRA Link 

https://polymesh.atlassian.net/browse/DA-65
https://polymesh.atlassian.net/browse/DA-66
https://polymesh.atlassian.net/browse/DA-67

### Changelog / Description 

Adds methods to
- addClaims
- editClaims
- revokeClaims

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/claims/claims.controller.spec.ts
+++ b/src/claims/claims.controller.spec.ts
@@ -42,7 +42,7 @@ describe('ClaimsController', () => {
   });
 
   describe('addClaims', () => {
-    it('should call addClaimsToDid method and return transaction data', async () => {
+    it('should call addClaimsOnDid method and return transaction data', async () => {
       mockClaimsService.addClaimsOnDid.mockResolvedValue({ ...txResult, result: undefined });
 
       const result = await controller.addClaims(mockPayload);
--- a/src/claims/claims.controller.spec.ts
+++ b/src/claims/claims.controller.spec.ts
@@ -1,0 +1,79 @@
+import { DeepMocked } from '@golevelup/ts-jest';
+import { Test } from '@nestjs/testing';
+import { ClaimType } from '@polymeshassociation/polymesh-sdk/types';
+
+import { ClaimsController } from '~/claims/claims.controller';
+import { ClaimsService } from '~/claims/claims.service';
+import { ModifyClaimsDto } from '~/claims/dto/modify-claims.dto';
+import { mockPolymeshLoggerProvider } from '~/logger/mock-polymesh-logger';
+import { testValues } from '~/test-utils/consts';
+import { mockClaimsServiceProvider } from '~/test-utils/service-mocks';
+
+const { did, txResult, signer } = testValues;
+
+describe('ClaimsController', () => {
+  let controller: ClaimsController;
+  let mockClaimsService: DeepMocked<ClaimsService>;
+
+  const mockPayload: ModifyClaimsDto = {
+    claims: [
+      {
+        target: did,
+        claim: {
+          type: ClaimType.Accredited,
+        },
+      },
+    ],
+    signer,
+  };
+
+  beforeEach(async () => {
+    const module = await Test.createTestingModule({
+      controllers: [ClaimsController],
+      providers: [mockClaimsServiceProvider, mockPolymeshLoggerProvider],
+    }).compile();
+
+    mockClaimsService = mockClaimsServiceProvider.useValue as DeepMocked<ClaimsService>;
+    controller = module.get<ClaimsController>(ClaimsController);
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
+
+  describe('addClaims', () => {
+    it('should call addClaimsToDid method and return transaction data', async () => {
+      mockClaimsService.addClaimsOnDid.mockResolvedValue({ ...txResult, result: undefined });
+
+      const result = await controller.addClaims(mockPayload);
+
+      expect(mockClaimsService.addClaimsOnDid).toHaveBeenCalledWith(mockPayload);
+
+      expect(result).toEqual({ ...txResult, results: undefined });
+    });
+  });
+
+  describe('editClaims', () => {
+    it('should call editClaimsOnDid method and return transaction data', async () => {
+      mockClaimsService.editClaimsOnDid.mockResolvedValue({ ...txResult, result: undefined });
+
+      const result = await controller.editClaims(mockPayload);
+
+      expect(mockClaimsService.editClaimsOnDid).toHaveBeenCalledWith(mockPayload);
+
+      expect(result).toEqual({ ...txResult, results: undefined });
+    });
+  });
+
+  describe('revokeClaims', () => {
+    it('should call revokeClaimsFromDid method and return transaction data', async () => {
+      mockClaimsService.revokeClaimsFromDid.mockResolvedValue({ ...txResult, result: undefined });
+
+      const result = await controller.revokeClaims(mockPayload);
+
+      expect(mockClaimsService.revokeClaimsFromDid).toHaveBeenCalledWith(mockPayload);
+
+      expect(result).toEqual({ ...txResult, results: undefined });
+    });
+  });
+});
--- a/src/claims/claims.controller.ts
+++ b/src/claims/claims.controller.ts
@@ -18,8 +18,8 @@ export class ClaimsController {
   }
 
   @ApiOperation({
-    summary: 'Get all Claims targeting an Identity',
-    description: 'This endpoint will provide a list of all the Claims made about an Identity',
+    summary: 'Add Claims targeting an Identity',
+    description: 'This endpoint will add Claims to an Identity',
   })
   @ApiTransactionResponse({
     description: 'Transaction response',
@@ -52,8 +52,8 @@ export class ClaimsController {
   }
 
   @ApiOperation({
-    summary: 'Get all Claims targeting an Identity',
-    description: 'This endpoint will provide a list of all the Claims made about an Identity',
+    summary: 'Remove provided Claims from an Identity',
+    description: 'This endpoint will remove Claims from an Identity',
   })
   @ApiTransactionResponse({
     description: 'Transaction response',
--- a/src/claims/claims.controller.ts
+++ b/src/claims/claims.controller.ts
@@ -27,7 +27,10 @@ export class ClaimsController {
     type: TransactionQueueModel,
   })
   @ApiTransactionFailedResponse({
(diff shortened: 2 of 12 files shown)


src/claims/claims.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+    private readonly claimsService: ClaimsService,
+    private readonly logger: PolymeshLogger
+  ) {
+    logger.setContext(ClaimsController.name);
+  }
+
+  @ApiOperation({
+    summary: 'Get all Claims targeting an Identity',
+    description: 'This endpoint will provide a list of all the Claims made about an Identity',
+  })
```
polymath-eric, line 23: Isn't this endpoint for creating a claim? This reads like its for `@Get`, not `@Post`

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+  @Post('claims/edit')
+  async editClaims(@Body() args: ModifyClaimsDto): Promise<TransactionResponseModel> {
+    const serviceResult = await this.claimsService.editClaimsOnDid(args);
+
+    return handleServiceResult(serviceResult);
+  }
+
+  @ApiOperation({
+    summary: 'Get all Claims targeting an Identity',
```
polymath-eric, line 55: This should say it removes claims

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.UNPROCESSABLE_ENTITY]: ['Account does not have the required roles or permissions'],
+  })
+  @Post('claims/remove')
+  async revokeClaims(@Body() args: ModifyClaimsDto): Promise<TransactionResponseModel> {
+    const serviceResult = await this.claimsService.revokeClaimsFromDid(args);
+
+    return handleServiceResult(serviceResult);
+  }
```
polymath-eric, line 69: Do we need a Get for claims?

There are the endpoint `@Get(':did/associated-claims')` and ` @Get(':did/issued-claims')` currently in the identities controller. I am not sure if they should get moved over to here though

polymath-eric: I was thinking we could move the controller method, but try to preserve the path. Maybe with taking the `target` and moving it to the path to be under identities.

Ideally we have the claims routes in one controller, and we don't make breaking changes.

If we do need a breaking change a path rename isn't that bad imo. Mark it  as such and we should get a major release.

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+  ) {
+    logger.setContext(ClaimsController.name);
+  }
+
+  @ApiOperation({
+    summary: 'Add Claims targeting an Identity',
+    description: 'This endpoint will add Claims to an Identity',
+  })
+  @ApiTransactionResponse({
+    description: 'Transaction response',
```
polymath-eric, line 25: I think the type needs to be specified so the documentation gets filled in.
```suggestion
    description: 'Transaction response',
     type: TransactionQueueModel,
```

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    return handleServiceResult(serviceResult);
+  }
+
+  @ApiOperation({
+    summary: 'Edit Claims targeting an Identity',
+    description: 'This endpoint will provide a list of all the Claims made about an Identity',
+  })
+  @ApiTransactionResponse({
+    description: 'Transaction response',
```
polymath-eric, line 42:
```suggestion
    description: 'Transaction response',
    type: TransactionQueueModel
```

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    return handleServiceResult(serviceResult);
+  }
+
+  @ApiOperation({
+    summary: 'Remove provided Claims from an Identity',
+    description: 'This endpoint will remove Claims from an Identity',
+  })
+  @ApiTransactionResponse({
+    description: 'Transaction response',
```
polymath-eric, line 59:
```suggestion
    description: 'Transaction response',
    type: TransactionQueueModel
```

src/claims/dto/modify-claims.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ApiProperty } from '@nestjs/swagger';
+import { Type } from 'class-transformer';
+import { IsNotEmpty, ValidateNested } from 'class-validator';
+
+import { ClaimTargetDto } from '~/claims/dto/claim-target.dto';
+import { TransactionBaseDto } from '~/common/dto/transaction-base-dto';
+
+export class ModifyClaimsDto extends TransactionBaseDto {
+  @ApiProperty({
+    description: 'The type of Claim. Note that different types require different fields',
```
polymath-eric, line 12:
```suggestion
    description: 'An array of Claims. Note that different types of Claims require different fields',
```

src/claims/claims.controller.ts
polymath-eric: If we aren’t sure it will be better I think its better to err on not making the breaking change. Its not like they are terribly out of place where they are, and I don't want people to have to update their stuff if we aren't sure it will benefit us.

I think there is benefit to having a stable API, and if we are breaking things we should have a strong reason imo.

That said, as far as breaking changes go, it will be small to move a route, and I don't think many people are using this endpoint.

polymath-eric (approved):

src/claims/claims.controller.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    mockClaimsService = mockClaimsServiceProvider.useValue as DeepMocked<ClaimsService>;
+    controller = module.get<ClaimsController>(ClaimsController);
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
+
+  describe('addClaims', () => {
+    it('should call addClaimsToDid method and return transaction data', async () => {
```
prashantasdeveloper, line 45:
```suggestion
    it('should call addClaimsOnDid method and return transaction data', async () => {
```

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+  @Post('add')
+  async addClaims(@Body() args: ModifyClaimsDto): Promise<TransactionResponseModel> {
+    const serviceResult = await this.claimsService.addClaimsOnDid(args);
+
+    return handleServiceResult(serviceResult);
+  }
+
+  @ApiOperation({
+    summary: 'Edit Claims targeting an Identity',
+    description: 'This endpoint will provide a list of all the Claims made about an Identity',
```
prashantasdeveloper, line 41: Description needs to be updated here.

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+  @ApiOperation({
+    summary: 'Add Claims targeting an Identity',
+    description: 'This endpoint will add Claims to an Identity',
+  })
+  @ApiTransactionResponse({
+    description: 'Transaction response',
+    type: TransactionQueueModel,
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.UNPROCESSABLE_ENTITY]: ['Account does not have the required roles or permissions'],
```
prashantasdeveloper, line 30: 'A target Identity cannot have CDD claims with different IDs' this should also be added

src/claims/claims.controller.ts
prashantasdeveloper: Make sure to specify that we can only update expiry of claim

src/claims/claims.controller.ts
prashantasdeveloper: Also need to add `Some of the supplied Identity IDs do not exist` for NOT_FOUND

src/claims/claims.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+  @ApiOperation({
+    summary: 'Remove provided Claims from an Identity',
+    description: 'This endpoint will remove Claims from an Identity',
+  })
+  @ApiTransactionResponse({
+    description: 'Transaction response',
+    type: TransactionQueueModel,
+  })
+  @ApiTransactionFailedResponse({
+    [HttpStatus.UNPROCESSABLE_ENTITY]: ['Account does not have the required roles or permissions'],
```
prashantasdeveloper, line 66: This also throws EntityInUse with this error - `Attempt to revoke Investor Uniqueness claims from investors with positive balance`

src/claims/dto/claim-target.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+import { ClaimDto } from '~/claims/dto/claim.dto';
+import { IsDid } from '~/common/decorators/validation';
+
+export class ClaimTargetDto {
+  @ApiProperty({
+    description: 'DID of the target Identity',
+    example: '0x0600000000000000000000000000000000000000000000000000000000000000',
+  })
+  @IsDid()
+  target: string;
```
prashantasdeveloper, line 15:
```suggestion
  readonly target: string;
```

src/claims/dto/claim-target.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  @ApiProperty({
+    description: 'The Claim to be added, modified or removed',
+  })
+  claim: ClaimDto;
```
prashantasdeveloper, line 20:
```suggestion
    description: 'The Claim to be added, modified or removed',
    type: ClaimDto,
  })
  @ValidateNested()
  @Type(() => ClaimDto)
  readonly claim: ClaimDto;
```

src/claims/dto/modify-claims.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Type } from 'class-transformer';
+import { IsNotEmpty, ValidateNested } from 'class-validator';
+
+import { ClaimTargetDto } from '~/claims/dto/claim-target.dto';
+import { TransactionBaseDto } from '~/common/dto/transaction-base-dto';
+
+export class ModifyClaimsDto extends TransactionBaseDto {
+  @ApiProperty({
+    description: 'An array of Claims. Note that different types of Claims require different fields',
+  })
```
prashantasdeveloper, line 13:
```suggestion
  @ApiProperty({
    description: 'An array of Claims. Note that different types of Claims require different fields',
    isArray: true,
    type: ClaimTargetDto,
  })
```

prashantasdeveloper (changes requested):
Make sure to add in all possible error responses available in the swagger failed responses decorator.

prashantasdeveloper (approved):

PR #153, opened 2022-12-20
feat: 🎸 add fireblocks signer

polymath-eric:
### JIRA Link 

[DA-469](https://polymesh.atlassian.net/browse/DA-469)

### Changelog / Description 

Allow operators of the REST API to configure it to use Fireblocks to approve transactions

### Checklist - 

- [X] New Feature ?
- [X] Updated swagger annotation (if API structure is changed) ?
- [X] Unit Test (if possible) ?
- [X] Updated the Readme.md (if required) ?


src/signing/signing.service.spec.ts
```diff
@@ -43,7 +54,7 @@ describe('LocalSigningService', () => {
 
   describe('initialize', () => {
     it('should call polymeshApi setSigningManager method', async () => {
-      await service.initialize({});
+      await service.initialize();
```
polymath-eric, line 57: Apparently default args are a branch, so falling back to no args adds test branch coverage.

src/signing/dto/signer-details.dto.ts
```diff
@@ -0,0 +1,13 @@
+/* istanbul ignore file */
+
+import { ApiProperty } from '@nestjs/swagger';
+import { IsString } from 'class-validator';
+
+export class SignerDetailsDto {
+  @ApiProperty({
+    description:
+      'The value to be used in the `signer` field. The format depends on what signer is being used',
```
prashantasdeveloper, line 9: Can we add multiple examples on for each type just for the sake of some clarity?

src/signing/signing.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    summary: 'Fetch signer details',
+    description: 'This endpoint provides information associated with a particular `signer`',
+  })
+  @ApiParam({
+    name: 'signer',
+    description:
+      'The value of the `signer` to fetch the address for. Note, the format depends on the signing manager the API is configured with',
+    type: 'string',
+    example: 'alice',
+  })
```
prashantasdeveloper, line 30: Since you have described the param here, you can remove the @ApiProperty from `SignerDetailsDto`

src/signing/signing.service.ts
```diff
@@ -101,3 +104,41 @@ export class VaultSigningService extends SigningService {
...
+     * Mainnet should use `595` as the coinType, otherwise it should be `1` to indicate a test net
+     * reference: https://github.com/satoshilabs/slips/blob/2a2f4c79508749f7e679a127d5a56da079b8d2d8/slip-0044.md?plain=1#L32
+     */
+    const coinType = this.signingManager.ss58Format === 12 ? 595 : 1;
+
+    const [accountId, change, accountIndex] = sections;
+
+    return [44, coinType, accountId, change || 0, accountIndex || 0];
+  }
+}
```
prashantasdeveloper, line 144: IMO this should be in its own file

prashantasdeveloper (approved):
Left some nit comments but overall LGTM

src/signing/signing.service.ts
```diff
@@ -101,3 +104,41 @@ export class VaultSigningService extends SigningService {
...
+      throw new AppValidationError(
+        'Fireblocks `signer` field should be 3 integers formatted like: `x-y-z`'
+      );
+    }
+
+    /**
+     * Mainnet should use `595` as the coinType, otherwise it should be `1` to indicate a test net
+     * reference: https://github.com/satoshilabs/slips/blob/2a2f4c79508749f7e679a127d5a56da079b8d2d8/slip-0044.md?plain=1#L32
+     */
+    const coinType = this.signingManager.ss58Format === 12 ? 595 : 1;
```
VictorVicente, line 138: Should we extract these kind of constants to a different file? Just to make them easier to change in the future if there is any need.

polymath-eric: I think if i break up the services into their own files and keep this grouped with the Fireblocks signer it will be more organized.

src/signing/signing.service.ts
polymath-eric: I can refactor each manager into its own file.

I was thinking of splitting them up myself, but figured I'd stick with the pattern.

src/signing/signing.service.ts
polymath-eric: Moved into `./util.ts` as a helper function

src/signing/dto/signer-details.dto.ts
polymath-eric: Added on ApiParam

PR #159, opened 2023-01-05
feat: 🎸 getClaimScopes

sansan:
### JIRA Link 

https://polymesh.atlassian.net/browse/DA-70

### Changelog / Description 

- get list of claim scopes for a did

### Checklist - 

- [x] New Feature ?


src/claims/models/claim-scope.model.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export class ClaimScopeModel {
+  @ApiProperty({
+    description: 'The scope that has been assigned to Identity',
+    example: {
+      type: 'Identity',
+      value: '0x6'.padEnd(66, '1a'),
+    },
+  })
+  readonly scope: Scope | null;
```
prashantasdeveloper, line 14: IIRC we have a ScopeModel which should be used here.

Also add nullable: true, define type field as well

src/claims/models/claim-scope.model.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(model: ClaimScopeModel) {
+    Object.assign(this, model);
+  }
+}
+
+export interface ClaimScope {
+  scope: Scope | null;
+  ticker?: string;
+}
```
prashantasdeveloper, line 30: This should be removed

prashantasdeveloper (changes requested):
Left some nit comments

src/claims/models/claim-scope.model.ts
```diff
@@ -1,30 +1,28 @@
...
   })
-  readonly scope: Scope | null;
+  @Type(() => ScopeModel)
+  readonly scope: ScopeModel | null;
 
   @ApiPropertyOptional({
+    type: 'string',
     description: 'The ticker to which the scope is valid for',
     example: 'TICKER',
+    nullable: true,
```
prashantasdeveloper, line 21: We only add nullable where field has `| null` I guess. So this should be removed. The `ApiPropertyOptional` already conveys the `?` in attribute

prashantasdeveloper (approved):

PR #163, opened 2023-01-09
feat: 🎸 modify portfolio name

sansan:
### JIRA Link 

https://polymesh.atlassian.net/browse/DA-76

### Changelog / Description 

- adds endpoint to modify portfolio name

### Checklist - 

- [x] New Feature ?


src/portfolios/portfolios.service.ts
```diff
@@ -78,4 +79,20 @@ export class PortfoliosService {
...
+    portfolioParams: PortfolioDto,
+    params: ModifyPortfolioDto
+  ): ServiceReturn<NumberedPortfolio> {
+    const { did, id } = portfolioParams;
+    const { base, args } = extractTxBase(params);
+
+    const portfolio = (await this.findOne(did, id)) as NumberedPortfolio;
+
+    if (!portfolio.modifyName) {
+      throw new BadRequestException('Default portfolio name cannot be modified');
```
polymath-eric, line 93:
```suggestion
      throw new AppValidationError('Default portfolio name cannot be modified');
```
There is a filter that will convert it for you so the behavior will be the same when used with the HTTP controller.

With something like REPL mode an HTTP error doesn't always make sense, which is what motivated me to make the new error type. 

The illustrative example is exposing the services over RPC.  It would be a leaky abstraction to have a different set of endpoints coupled to HTTP details.

polymath-eric: I admit its bit nitty, since I think any one using a REPL knows what HTTP codes are, but I think the cleaner abstraction is worth it.

src/portfolios/portfolios.service.spec.ts
```diff
@@ -242,4 +242,51 @@ describe('PortfoliosService', () => {
...
+        tag: TxTags.portfolio.RenamePortfolio,
+      };
+      const mockTransaction = new MockTransaction(transaction);
+
+      const mockIdentity = new MockIdentity();
+      const modifyName = jest.fn();
+
+      modifyName.mockReturnValue(mockTransaction);
+
+      const mockPortfolio = {
```
prashantasdeveloper, line 261: This should use MockPortfolio from `mocks.ts`

src/portfolios/portfolios.service.spec.ts
```diff
@@ -242,4 +242,51 @@ describe('PortfoliosService', () => {
...
+        name: 'FOLIO-1',
+      };
+
+      const result = await service.updatePortfolioName(portfolio, body);
+      expect(result).toEqual({
+        result: mockPortfolio,
+        transactions: [mockTransaction],
+      });
+    });
+  });
```
prashantasdeveloper, line 291: Add in test case that throws error for default portfolio

prashantasdeveloper (changes requested):

src/portfolios/portfolios.service.ts
```diff
@@ -78,4 +80,20 @@ export class PortfoliosService {
...
     );
   }
+
+  public async updatePortfolioName(
+    portfolioParams: PortfolioDto,
+    params: ModifyPortfolioDto
+  ): ServiceReturn<NumberedPortfolio> {
+    const { did, id } = portfolioParams;
+
+    if (id.toNumber() <= 0) {
```
polymath-eric, line 90:
```suggestion
    if (id.lte(0)) {
```
(nit) BigNumber has compare functions to use

prashantasdeveloper (approved):

src/portfolios/portfolios.service.ts
```diff
@@ -78,4 +80,20 @@ export class PortfoliosService {
...
+    params: ModifyPortfolioDto
+  ): ServiceReturn<NumberedPortfolio> {
+    const { did, id } = portfolioParams;
+
+    if (id.lte(0)) {
+      throw new AppValidationError('Default portfolio name cannot be modified');
+    }
+
+    const { base, args } = extractTxBase(params);
+    const portfolio = (await this.findOne(did, id)) as NumberedPortfolio;
```
polymath-eric, line 95: (optional)

We can use function overloading to clean up this assertion. We use this pattern in the SDK in a few places.

```ts
 public async findOne(did: string): Promise<DefaultPortfolio>;
  public async findOne(did: string, portfolioId: BigNumber): Promise<NumberedPortfolio>;
  public async findOne(
    did: string,
    portfolioId?: BigNumber
  ): Promise<DefaultPortfolio | NumberedPortfolio> {
```

Only down side is moveAssets needs to get modified like:

```
    const fromId = toPortfolioId(from);
    const fromPortfolio = fromId ? await this.findOne(owner, fromId) : await this.findOne(owner);
```

but it gives us stronger types and prevents the assertion here.

polymath-eric, line 95: Update `findOne` with this signature and you won't need to cast. Also mentioned [here](https://github.com/PolymeshAssociation/polymesh-rest-api/pull/163/files#r1081448512)

```
public async findOne(did: string): Promise<DefaultPortfolio>;
  public async findOne(did: string, portfolioId: BigNumber): Promise<NumberedPortfolio>;
  public async findOne(
    did: string,
    portfolioId?: BigNumber
  ): Promise<DefaultPortfolio | NumberedPortfolio> {
```

polymath-eric: Since its used in a few places I think the clean up is worth it. 

I don't like how type assertion looks with `await`, and TS provides a feature to solve this.

polymath-eric (approved):

PR #168, opened 2023-01-18
Revert "chore: 🤖 publish all releases, not just "published" ones"

polymath-eric:
Reverts PolymeshAssociation/polymesh-rest-api#164

It seems like this didn't work, and made it so unmarking the release didn't trigger the pipeline.

I think unchecking the box for alpha isn't too bad, so I think this reversion will be OK for now.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - v*
+  release:
+    type: [published]
 
 jobs:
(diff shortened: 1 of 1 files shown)


prashantasdeveloper (approved):

VictorVicente (approved):

polymath-eric:
I think the issue is because we have two actions (release + docker)and Github prevents an action from triggering another action to prevent infinite loops.

PR #181, opened 2023-03-02
feat: 🎸 Export swagger json in release assets

prashantasdeveloper:
### JIRA Link 

DA-501

### Changelog / Description 

On release of any new version - 
* Writes swagger document to `polymesh-rest-api-swagger-spec.json` file 
* Updates `package.json` with apt version
* Updates `src/main.ts` to the same version to help identify the version when someone spins up their own REST API instance

### Checklist - 

- [x] New Feature ?


release.config.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
     [
       '@semantic-release/github',
       {
-        assets: ['CHANGELOG.md'],
+        assets: ['CHANGELOG.md', 'swagger-spec.json'],
```
prashantasdeveloper, line 19: Theoretically this should do the trick, but I was not able to test this locally. Will have to merge a PR to get this tested. Any other way to do this @polymath-eric @VictorVicente ?

src/main.ts
```diff
@@ -59,7 +60,7 @@ async function bootstrap(): Promise<void> {
   if (isApiKeyStrategyConfigured) {
     document.security = [{ api_key: [] }]; // Apply the API key globally to all operations
   }
-
+  fs.writeFileSync('./swagger-spec.json', JSON.stringify(document));
```
polymath-eric, line 63: I was thinking along the lines of having a separate script that would run in the actions and not touch a developers HD.

This is probably OK and is less work. As long as we commit the file we should be good. I can't see it getting out of date when its updated on start

polymath-eric (approved):
The way I was thinking was more work, but I think what you have is good

src/main.ts
VictorVicente: But a script might be a cleaner solution IMO. It will ensure to upload the corresponding swagger file for each build rather than depend to commit it.

polymath-eric: @prashantasdeveloper you can look at the `repl.ts` + package.json script using it as an example

VictorVicente (approved):

src/commands/write-swagger.ts
```diff
@@ -0,0 +1,20 @@
+import { NestFactory } from '@nestjs/core';
+import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
+import { writeFileSync } from 'fs';
+
+import { AppModule } from '~/app.module';
+
+const writeSwaggerSpec = async (): Promise<void> => {
+  const app = await NestFactory.create(AppModule, { logger: false });
```
polymath-eric, line 8: @prashantasdeveloper wether this is a test module, or a normal one, this still wants to call out when we configure the module. We had the issue of this getting masked from local `.env` files

I don't want to call out to a non local node in CI. The easiest fix would be to start a docker container, but that seems overkill when its not actually needed.

Our mocks are dependent on having jest. Its tempting, but seems like tech debt to exploit DI with a mock though.

I am no sure on a clean way to solve this and don't want to get more side tracked from CDD at this point. No urgency on this PR, so I can revisit later this week. I am certainly appreciating the simplicity of the write + commit approach, but that has draw backs when other people are running this imo.

src/main.ts
polymath-eric: I am certainly appreciating the simplicity now.

For this to run in CI we need to figure out a way to not need a Polymesh node without making a mess out of our config

package.json
```diff
@@ -15,10 +15,11 @@
     "commit": "npx git-cz",
     "format": "prettier-eslint --write \"src/**/*.ts\" \"test/**/*.ts\"",
     "start": "nest start",
-    "start:repl": "yarn start --entryFile repl",
     "start:dev": "nest start --watch",
     "start:debug": "nest start --debug --watch",
     "start:prod": "NODE_ENV='production' node dist/main",
+    "start:repl": "yarn start --entryFile commands/repl",
+    "generate:swagger": "yarn start --entryFile commands/write-swagger",
```
polymath-eric, line 22:
```suggestion
    "generate:swagger": "POLYMESH_NODE_URL: 'wss://testnet-rpc.polymesh.live' yarn start --entryFile commands/write-swagger",
```

I am thinking this is probably the best stop gap. Downside being testnet being down would block a release, but we can always rerun / manually add if something happens.

prashantasdeveloper: For now I propose, lets stick to this to verify that the process we created works and once that is done we can setup the chain as well later

package.json
```diff
@@ -73,6 +74,7 @@
     "@nestjs/schematics": "^9.0.3",
     "@nestjs/testing": "^9.1.6",
     "@semantic-release/changelog": "^6.0.1",
+    "@semantic-release/git": "^10.0.1",
```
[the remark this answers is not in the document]
prashantasdeveloper: I need this to commit the package.json file and src/main.ts

prepareRelease.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+declare nextVersion=$1
+
+sed -i.bak -e "s/.setVersion('.*')/.setVersion('$nextVersion')/g" src/main.ts
```
prashantasdeveloper, line 5: This should update the version in the released tags. (Remembered that @polymath-eric wanted this wrt OTC, maybe it's not needed but I would like to keep this as well so that anyone spinning a new instance would get to know easily)

prepareRelease.sh
```diff
@@ -0,0 +1,8 @@
...
+rm src/main.ts.bak
+
+SWAGGER_VERSION=$nextVersion POLYMESH_NODE_URL='wss://testnet-rpc.polymesh.live' yarn generate:swagger > /dev/null 2>&1
```
prashantasdeveloper, line 8: I have stuck with testnet RPC for now. We can later think of replace this with a local chain @polymath-eric

release.config.js
```diff
@@ -13,6 +13,25 @@ module.exports = {
...
+      {
+        npmPublish: false,
+      },
+    ],
+    [
+      '@semantic-release/git',
+      {
+        assets: ['package.json', 'src/main.ts'],
+      },
+    ],
```
prashantasdeveloper, line 34: 💡 The idea here is - 
- `prepareCmd` will exec will trigger our script with the nextRelease version.
- @semantic-release/npm will update the package.json with apt version. Since we don't want to publish to npm we set it to false
- @semantic-release/git will help us commit the changes in assets file list

prepareRelease.sh
polymath-eric: (nit) I am always a bit worried when making raw text changes like this, especially when its changing code

Instead of: `.setVersion('1.0');` what about `.setVersion(REST_API_VERSION)`

and then target that variable to replace where its the top of the file (or even its own dedicated file)

That said, what you have is probably OK, but we don't have any release stopping test coverage on main.ts in either.

polymath-eric (approved):
LGTM, left a nitty comment around `sed`

VictorVicente (approved):

PR #182, opened 2023-03-06
refactor: 💡 use nest to run swagger generation instead of jest

polymath-eric:
Jest is meant to be a test runner, and not a general purpose task runner.

I think its cleaner to use just NestJS.

I tried getting it to write to stdout, but extra stuff got added in. the `logger` false keep it pretty quite though.


.gitignore
```diff
@@ -1,6 +1,7 @@
 # compiled output
 /dist
 /node_modules
+polymesh-rest-api-swagger-spec.json
```
polymath-eric, line 4: This maybe could go in `/dist`, but not sure how that affects the pipe line.

prashantasdeveloper (approved):

src/developer-testing/developer-testing.module.ts
```diff
@@ -14,8 +14,7 @@ export class DeveloperTestingModule {
   static register(): DynamicModule {
     const controllers = [];
 
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const DEVELOPER_UTILS: boolean = JSON.parse(process.env.DEVELOPER_UTILS!);
+    const DEVELOPER_UTILS: boolean = JSON.parse(`${!!process.env.DEVELOPER_UTILS}`);
```
polymath-eric, line 17:
```ts
const DEVELOPER_UTILS: boolean = JSON.parse(process.env.DEVELOPER_UTILS!);
                                          ^
SyntaxError: Unexpected token u in JSON at position 0
```

TypeScript was right again

polymath-eric:
seems like I war relying on having a .env defined. working on the errors

.gitignore
polymath-eric: This should NOT go into `/dist` its ~300KB and would be bloat in an image / package.

PR #184, opened 2023-03-12
feat: 🎸 Add endpoint to get transaction details by hash

prashantasdeveloper:
### JIRA Link 

DA-530

### Changelog / Description 

Add endpoint to get transaction details by hash

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/transactions/dto/transaction-hash-params.dto.ts
+++ b/src/transactions/dto/transaction-hash-params.dto.ts
@@ -1,8 +1,11 @@
 /* istanbul ignore file */
 
-import { IsHexadecimal } from 'class-validator';
+import { IsHexadecimal, MaxLength } from 'class-validator';
+
+import { MAX_HASH_LENGTH } from '~/transactions/transactions.consts';
 
 export class TransactionHashParamsDto {
   @IsHexadecimal()
+  @MaxLength(MAX_HASH_LENGTH)
   readonly hash: string;
 }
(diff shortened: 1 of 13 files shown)


src/transactions/transactions.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+    @Res() res: Response
+  ): Promise<void> {
+    const result = await this.networkService.getTransactionByHash(hash);
+
+    if (result) {
+      res.status(HttpStatus.OK).json(new ExtrinsicDetailsModel(result));
+    } else {
+      res.status(HttpStatus.NO_CONTENT).send({});
+    }
+  }
```
polymath-eric, line 41: This will have the same effect, Nest defaults to no content if it has nothing to render
```suggestion
 ): Promise<ExtrinsicDetailsModel | void> {
    const result = await this.networkService.getTransactionByHash(hash);

    if (result) {
      return new ExtrinsicDetailsModel(result);
    }
  }
```

polymath-eric (approved):

src/transactions/dto/transaction-hash-params.dto.ts
```diff
@@ -0,0 +1,8 @@
+/* istanbul ignore file */
+
+import { IsHexadecimal } from 'class-validator';
+
+export class TransactionHashParamsDto {
+  @IsHexadecimal()
```
polymath-eric, line 6: We can do a length check here as well since the transaction hash should be a fixed size.

VictorVicente: @prashantasdeveloper could you add the length check so we can merge it as well?

polymath-eric (approved):

PR #187, opened 2023-03-21
fix: 🐛 bump dependencies

VictorVicente:
### JIRA Link 

DA-541

### Changelog / Description 

Bump dependencies to solve security alertes raised by dependabot

### Checklist - 

- [ ] New Feature ?


polymath-eric (approved):

polymath-eric:
Looks like the enum validation message is better now.

This is a nice case for "toMatchSnapshot` or `toMatchInlineSnapshot` for the fix to be "run with -u", but its good our tests caught that change imo.

PR #191, opened 2023-04-03
chore: 🤖 upgrade to node V18 runtime

polymath-eric:
makes handleSdkError return an error and have the caller throw to provide more flexibility. Adds in global unhandledPromiseRejection handler to ensure compatibility with V18

✅ Closes: [DA-546](https://polymesh.atlassian.net/browse/DA-546)

[DA-546]: https://polymesh.atlassian.net/browse/DA-546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/accounts/accounts.service.ts
+++ b/src/accounts/accounts.service.ts
@@ -28,7 +28,9 @@ export class AccountsService {
     const {
       polymeshService: { polymeshApi },
     } = this;
-    return await polymeshApi.accountManagement.getAccount({ address }).catch(handleSdkError);
+    return await polymeshApi.accountManagement.getAccount({ address }).catch(error => {
+      throw handleSdkError(error);
+    });
   }
 
   public async getAccountBalance(account: string): Promise<AccountBalance> {
@@ -64,7 +66,9 @@ export class AccountsService {
 
   public async getPermissions(address: string): Promise<Permissions> {
     const account = await this.findOne(address);
-    return await account.getPermissions().catch(handleSdkError);
+    return await account.getPermissions().catch(error => {
+      throw handleSdkError(error);
+    });
   }
 
   public async freezeSecondaryAccounts(
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,6 +1,6 @@
 /* istanbul ignore file */
 
-import { ClassSerializerInterceptor, ValidationPipe } from '@nestjs/common';
+import { ClassSerializerInterceptor, Logger, ValidationPipe } from '@nestjs/common';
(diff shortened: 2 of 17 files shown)


src/transactions/transactions.util.ts
```diff
@@ -140,32 +145,37 @@ export async function processTransaction<
...
     };
   } catch (err) {
-    handleSdkError(err);
+    throw handleSdkError(err);
   }
 }
 
-export function handleSdkError(err: unknown): never {
+export function handleSdkError(err: unknown): AppError {
+  if (isAppError(err)) {
```
polymath-eric, line 153: This is important for when errors get re thrown. It didn't solve the unhandled rejection promise I was looking for, but it did fix the "pause" already paused asset.

Although not 100% necessary for this to return, I think its much more flexible to give the choice to the caller compared to throw

src/accounts/accounts.service.ts
```diff
@@ -28,7 +28,9 @@ export class AccountsService {
     const {
       polymeshService: { polymeshApi },
     } = this;
-    return await polymeshApi.accountManagement.getAccount({ address }).catch(handleSdkError);
+    return await polymeshApi.accountManagement.getAccount({ address }).catch(error => {
+      throw handleSdkError(error);
```
polymath-eric, line 32: I was the one who suggested to use the short form of no function declaration, but without the arrow function the "this" in handleSdk can get messed up.

There wasn't an issue before, but it could cause an unexpected bug if handleSdkError started using "this" without this change. I think its safer to use the arrow function, even if its looks like visual clutter due to JS being JS.

src/main.ts
```diff
@@ -13,6 +13,13 @@ import { LoggingInterceptor } from '~/common/interceptors/logging.interceptor';
 import { WebhookResponseCodeInterceptor } from '~/common/interceptors/webhook-response-code.interceptor';
 import { PolymeshLogger } from '~/logger/polymesh-logger.service';
 
+// This service was originally designed with node v14, this ensures a backwards compatible run time
+// Ideally we wouldn't need this function, but I am unable to find the cause when submitting SDK transactions that fail validation
+const unhandledRejectionLogger = new Logger('UnhandledPromise');
```
polymath-eric, line 18: I can't figure out why the service crashes without this. This makes the service work like before though without breaking when called with the wrong node flag.

VictorVicente (approved):

PR #221, opened 2023-11-23
feat: 🎸 get all registered custom claims

sansan:
### JIRA Link 

https://polymesh.atlassian.net/browse/DA-941

### Changelog / Description 

- adds endpoint to retrieve all registered CustomClaimTypes
- extracts getCustomClaimTypeByName to a separate endpoint
- extracts getCustomClaimTypeById to a separate endpoint 

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/claims/claims.controller.ts
+++ b/src/claims/claims.controller.ts
@@ -10,12 +10,11 @@ import {
 } from '@nestjs/common';
 import { ApiNotFoundResponse, ApiOperation, ApiParam, ApiTags } from '@nestjs/swagger';
 import { BigNumber } from '@polymeshassociation/polymesh-sdk';
+import { CustomClaimType } from '@polymeshassociation/polymesh-sdk/types';
 
 import { ClaimsService } from '~/claims/claims.service';
-import {
-  GetCustomClaimTypeByIdDto,
-  GetCustomClaimTypeByNameDto,
-} from '~/claims/dto/get-custom-claim-type.dto';
+import { GetCustomClaimTypePipe } from '~/claims/decorators/get-custom-claim-type.pipe';
+import { GetCustomClaimTypeDto } from '~/claims/dto/get-custom-claim-type.dto';
 import { GetCustomClaimTypesDto } from '~/claims/dto/get-custom-claim-types.dto';
 import { ModifyClaimsDto } from '~/claims/dto/modify-claims.dto';
 import { RegisterCustomClaimTypeDto } from '~/claims/dto/register-custom-claim-type.dto';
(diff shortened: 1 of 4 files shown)


src/claims/claims.controller.ts
```diff
@@ -105,18 +121,58 @@ export class ClaimsController {
   }
 
   @ApiOperation({
-    summary: 'Get CustomClaimType',
-    description:
-      'This endpoint will retrieve the custom claim type by id or name. <br /> ID or name needs to be provided. <br/> If both are provided will fetch using ID',
+    summary: 'Get CustomClaimTypes',
+    description: 'This endpoint fetches CustomClaimTypes that have been registered."',
```
polymath-eric, line 125:
```suggestion
    description: 'This endpoint fetches CustomClaimTypes that have been registered"',
```

(nit) I don't think we put in trailing period usually

src/claims/claims.controller.ts
```diff
@@ -105,18 +121,58 @@ export class ClaimsController {
...
     description: 'The ID of the CustomClaimType',
     type: 'string',
     required: false,
     example: '1',
   })
-  @ApiQuery({
+  @ApiNotFoundResponse({
+    description: 'The CustomClaimType does not exist',
+  })
+  @Get('custom-claim-types/id/:id')
```
polymath-eric, line 158:
```suggestion
  @Get('custom-claim-types/:id')
```

Seems weird to have `/id` in the route to me. I see its to distringuish between name though.

If name is always unique I think we could have a single endpoint that would return details for either.

```
  @Get('custom-claim-types/:id')
  
  if(id instanceof BigNumber) { 
    this.claimsService.getCustomClaimTypeById(id);
   else {
     this.claimsService.getCustomClaimTypeByName(name);
   }
  ```
  
  This might make the validators a bit tricky though.

**Review by polymath-eric — COMMENTED** [review]
The /id and /name routes don't seem unconventional path wise. Not sure if we can combine them or use a query param instead.

Kind of nitty so I could be convinced to leave it as it

File: package.json [hunk 3]
```diff
@@ -44,11 +44,11 @@
...
     "@nestjs/swagger": "^6.2.1",
     "@nestjs/typeorm": "^9.0.1",
+    "@polymeshassociation/local-signing-manager": "^3.0.1",
+    "@polymeshassociation/signing-manager-types": "^3.0.0",
     "@polymeshassociation/fireblocks-signing-manager": "^2.2.0",
     "@polymeshassociation/hashicorp-vault-signing-manager": "^2.1.0",
-    "@polymeshassociation/local-signing-manager": "^2.1.0",
-    "@polymeshassociation/polymesh-sdk": "23.0.0-alpha.27",
-    "@polymeshassociation/signing-manager-types": "^2.1.0",
+    "@polymeshassociation/polymesh-sdk": "23.0.0-alpha.34",
```
polymath-eric, line 51: I just merged v23 release. When that gets published this can get bumped to `"^23.0.0"`

PR #234, opened 2024-01-10
feat: 🎸 config artemis MQ explictly

polymath-eric:
### JIRA Link 

[DA-1001](https://polymesh.atlassian.net/browse/DA-1001)

### Changelog / Description 

Configures ArtemisMQ explicitly + adds graceful shutdown and eliminates redundant connections to Artemis

### Checklist - 

- [ ] New Feature ?
- [X] Unit Test (if possible) ?


Procfile
```diff
@@ -1 +0,0 @@
-web: npm run start:prod
```
polymath-eric, line 1: this file is for heroku which we never use.

compose/broker.xml
```diff
@@ -0,0 +1,145 @@
+<?xml version='1.0'?>
```
polymath-eric, line 1: I got this base of this file by looking at the auto generated one in a blank service and then tweaking it.

I left lots of the comments in it, but also removed unnecessary details. Values not specified should fallback on their default

compose/broker.xml
```diff
@@ -0,0 +1,145 @@
...
+         </address>
+
+         <address name="ExpiryQueue">
+            <anycast>
+               <queue name="ExpiryQueue" />
+            </anycast>
+         </address>
+
+         <address name="Requests">
+            <anycast>
```
polymath-eric, line 101: I tried to make this `multicast` so I could keep the old submitter to record the body. However I couldn't get the model to work with Artemis via AMQP. What I was trying to do might have been an anti pattern though.

For jobs to be done, we should use multicast, where as notifications of things that happened should be on multicast, like finalizations.

compose/broker.xml
```diff
@@ -0,0 +1,145 @@
...
+        </address>
+
+        <address name="EventsLog">
+          <anycast>
+             <queue name="EventsLog"></queue>
+          </anycast>
+        </address>
+      </addresses>
+
+      <diverts>
```
polymath-eric, line 124: Diverts let the event recorder monitor an otherwise single consumable queue without needing multicast.

We could still register it to different multicast queues if we really need, but once a transaction is finalized details will end up in SubQuery and the chain, so I don't want to add any more events just yet.

compose/init-db.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
+    SELECT 'CREATE DATABASE rest'
+    WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = 'rest')\gexec
```
polymath-eric, line 6: This is so we can have one postgres in the compose file, and use it for SQ + REST api.

Without this file you would have to manually make the DB. Its mapped to be in an "auto run" dir in the container

src/artemis/artemis.service.ts
```diff
@@ -14,27 +14,70 @@ import {
...
+        this.logger.error(`error closing artemis connection: ${result.reason}`);
+      } else {
+        successfulCloses += 1;
+      }
+    }
+    this.logger.debug(`successfully closed ${successfulCloses} connections`);
+
+    const connection = await this.getConnection();
+
+    await connection.close();
```
polymath-eric, line 65: not cleaning up can give warnings on the Artemis broker

src/artemis/artemis.service.ts
```diff
@@ -112,28 +156,50 @@ export class ArtemisService {
     });
   }
 
-  private async setupQueue(topicName: TopicName): Promise<QueueEntry> {
-    const entry = this.queueStore[topicName];
+  private async getConnection(): Promise<Connection> {
+    if (!this.connectionPromise) {
```
polymath-eric, line 160: I thought I _might_ have been getting a race, so added this kind of cache.

I don't think I was, but this way seems safer to make sure we only have one connection

src/datastore/postgres/migrations/1704924533614-offline.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class Offline1704924533614 implements MigrationInterface {
+  name = 'Offline1704924533614';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query('ALTER TABLE "offline_event" DROP COLUMN "topicName"');
```
polymath-eric, line 7: Diverts loose the original address information.

We could add it as a property into the message if we really wanted it though

VictorVicente (approved):

PR #249, opened 2024-03-04
feat: 🎸 Add APIs using SQ SDK methods

prashantasdeveloper:
### JIRA Link 

DA-1110, DA-1070

### Changelog / Description 

Adds API to get held confidential assets, get involved confidential transactions

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,9 +1,15 @@
 {
   "compilerOptions": {
-    "outDir": "dist",    
+    "outDir": "dist",
     "baseUrl": ".",
     "paths": {
       "~/*": ["src/*"],
+      "@polymeshassociation/polymesh-sdk": [
+        "node_modules/@polymeshassociation/confidential-polymesh-sdk"
+      ],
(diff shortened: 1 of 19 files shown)


tsconfig.json
```diff
@@ -1,9 +1,15 @@
...
-    "outDir": "dist",    
+    "outDir": "dist",
     "baseUrl": ".",
     "paths": {
       "~/*": ["src/*"],
+      "@polymeshassociation/polymesh-sdk": [
+        "node_modules/@polymeshassociation/confidential-polymesh-sdk"
+      ],
+      "@polymeshassociation/polymesh-sdk/*": [
+        "node_modules/@polymeshassociation/confidential-polymesh-sdk/*"
```
polymath-eric, line 11: seems like jest isn't able to pick this up.

maybe we need to do find + replace across the code base instead? Might be annoying to have to update every import line when syncing with upstream though

prashantasdeveloper: I need to added in jest as well

PR #293, opened 2024-10-21
feat: 🎸 Add support for Asset IDs 

prashantasdeveloper:
### JIRA Link 

DA-1301

### Changelog / Description 

Adds support for Asset IDs
- All the asset related endpoints (starting with `assets/` or associated
with `assets` tag) now also accept asset ID as permissible value along with ticker for
the corresponding path param from 7.x chain.
- The `ticker` property of `CheckpointScheduleModel` has been renamed to `asset`.
- The query param `ticker` has been renamed to `asset` for the endpoint
GET `portfolios/identities/:did/portfolios/:id/transactions`.
- The property `ticker` has been renamed to `asset` in `PortfolioMovementDto`.
This affects the endpoint POST `portfolios/identities/:did/portfolios/move-assets`.
- The endpoint POST `settlements/instructions/create` now accepts asset ID as well
for on chain legs, as possible value for `asset` body params.
- The endpoint GET `settlements/leg-validations` now accepts asset ID as well,
as possible value for `asset` body param from 7.x chain.
- The `asset` property of `CorporateActionModel`, `PreApprovedModel`,
`MetadataEntryModel` now return asset ID values from 7.x chain.

### Checklist - 

- [x] New Feature ?

### DIFF
--- a/src/assets/assets.controller.ts
+++ b/src/assets/assets.controller.ts
@@ -13,13 +13,13 @@ import {
 
 import { AssetsService } from '~/assets/assets.service';
 import { createAssetDetailsModel } from '~/assets/assets.util';
+import { AssetParamsDto } from '~/assets/dto/asset-params.dto';
 import { ControllerTransferDto } from '~/assets/dto/controller-transfer.dto';
 import { CreateAssetDto } from '~/assets/dto/create-asset.dto';
 import { IssueDto } from '~/assets/dto/issue.dto';
 import { RedeemTokensDto } from '~/assets/dto/redeem-tokens.dto';
 import { RequiredMediatorsDto } from '~/assets/dto/required-mediators.dto';
 import { SetAssetDocumentsDto } from '~/assets/dto/set-asset-documents.dto';
-import { TickerParamsDto } from '~/assets/dto/ticker-params.dto';
 import { AgentOperationModel } from '~/assets/models/agent-operation.model';
 import { AssetDetailsModel } from '~/assets/models/asset-details.model';
 import { AssetDocumentModel } from '~/assets/models/asset-document.model';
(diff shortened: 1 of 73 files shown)


src/assets/assets.controller.ts
```diff
@@ -91,10 +91,10 @@ export class AssetsController {
       'This endpoint will provide the list of Asset holders along with their current balance',
   })
   @ApiParam({
-    name: 'ticker',
-    description: 'The ticker of the Asset whose holders are to be fetched',
+    name: 'asset',
+    description: 'The Asset (Ticker/Asset ID) whose holders are to be fetched',
     type: 'string',
-    example: 'TICKER',
+    example: '0xa3616b82e8e1080aedc952ea28b9db8b',
```
polymath-eric, line 97: I thought these were UUIDs now. Should this be formatted with the dashes?

src/assets/assets.service.ts
```diff
@@ -34,18 +35,30 @@ export class AssetsService {
     private readonly transactionsService: TransactionsService
   ) {}
 
-  public async findOne(ticker: string): Promise<Asset> {
-    return await this.polymeshService.polymeshApi.assets.getAsset({ ticker }).catch(error => {
+  public async findOne(asset: string): Promise<Asset> {
+    let getAssetPromise;
+    if (asset.length === ASSET_ID_LENGTH) {
```
polymath-eric, line 40: There should be a function `isAssetId` that captures the actual logic.

```ts
const assetIdRegex = /0x[0-9A-F]{32}/;
const isAssetId = (id: string): boolean => {
  return assetIdRegex.test(id);
};
```

Then the code can be shared with the validator and we always perform the most exact check we can.

src/assets/assets.service.ts
```diff
@@ -34,18 +35,30 @@ export class AssetsService {
...
   }
 
-  public async findFungible(ticker: string): Promise<FungibleAsset> {
-    return await this.polymeshService.polymeshApi.assets
-      .getFungibleAsset({ ticker })
-      .catch(error => {
-        throw handleSdkError(error);
+  public async findFungible(asset: string): Promise<FungibleAsset> {
+    let getAssetPromise;
+    if (asset.length === ASSET_ID_LENGTH) {
```
polymath-eric, line 52: `isAssetId(asset)`

src/common/decorators/validation.ts
```diff
@@ -42,6 +42,31 @@ export function IsTicker(validationOptions?: ValidationOptions) {
...
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'isAsset',
+      target: object.constructor,
+      propertyName,
+      options: validationOptions,
+      validator: {
+        validate(value: string) {
+          if (value.startsWith('0x')) {
```
polymath-eric, line 55:
```suggestion
          if(isAssetId(value) {
```

polymath-eric (changes requested):
isAssetId logic needs to get its own function. If that gets changed to UUID format then we should only need to update it there. Where ever there is a length check in the code it gets coupled to the current format logic and isn't obvious that it needs to be updated.

src/assets/assets.controller.ts
prashantasdeveloper: Leaving this as it is for now. Once the final decision is taken on the format to be displayed we can update all the instance

src/common/decorators/validation.ts
```diff
@@ -42,6 +42,33 @@ export function IsTicker(validationOptions?: ValidationOptions) {
   );
 }
 
+const assetIdRegex = /0x[0-9A-F]{32}/;
```
polymath-eric, line 45:
```suggestion
const assetIdRegex = /0x[0-9a-fA-F]{32}/;
```

I missed lower case support in my initial comment 😅

PR #348, opened 2025-09-25
feat: 🎸 get protocol fees

sansan:
### JIRA Link 

https://polymesh.atlassian.net/browse/DA-1549

### Changelog / Description 

adds endpoint to get protocol fees

### Checklist - 

- [x] New Feature 
- [x] Updated swagger annotation (if API structure is changed) 
- [x] Unit Test (if possible)

### DIFF
--- a/src/network/dto/protocol-fees-query.dto.ts
+++ b/src/network/dto/protocol-fees-query.dto.ts
@@ -1,0 +1,31 @@
+/* istanbul ignore file */
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { TxTag, TxTags } from '@polymeshassociation/polymesh-sdk/types';
+import { Transform } from 'class-transformer';
+import { ArrayNotEmpty, IsArray, IsOptional, IsString } from 'class-validator';
+
+import { IsTxTag } from '~/common/decorators/validation';
+import { getTxTags, toArray } from '~/common/utils';
+
+export class ProtocolFeesQueryDto {
+  @ApiProperty({
+    description: 'Transaction tags whose protocol fees should be returned',
+    isArray: true,
+    enum: getTxTags(),
+    example: [TxTags.asset.CreateAsset],
+  })
+  @Transform(({ value }) => toArray(value))
+  @IsArray()
+  @ArrayNotEmpty()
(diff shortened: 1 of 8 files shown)


src/network/dto/protocol-fees-query.dto.ts
```diff
@@ -0,0 +1,46 @@
...
+  @Transform(({ value }) => toArray(value))
+  @IsArray()
+  @ArrayNotEmpty()
+  @IsTxTag({ each: true })
+  readonly tags: TxTag[];
+
+  @ApiPropertyOptional({
+    description: 'Optional block hash to query historic protocol fees',
+    type: 'string',
+    example: '0x1234abcd',
```
polymath-eric, line 41: I think the example should be a full hash

src/network/dto/protocol-fees-query.dto.ts
```diff
@@ -0,0 +1,46 @@
+/* istanbul ignore file */
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { TxTag, TxTags } from '@polymeshassociation/polymesh-sdk/types';
+import { Transform } from 'class-transformer';
+import { ArrayNotEmpty, IsArray, IsOptional, IsString } from 'class-validator';
+
+import { IsTxTag } from '~/common/decorators/validation';
+import { getTxTags } from '~/common/utils';
+
+const toArray = (value: unknown): string[] | undefined => {
```
polymath-eric, line 10: This seems like a function that should be generic. I guess its fine to leave, but feels like something that should be in `~/common/utils`

src/network/network.service.ts
```diff
@@ -29,6 +30,12 @@ export class NetworkService {
     return this.polymeshService.polymeshApi.network.getTreasuryBalance();
   }
 
+  public async getProtocolFees(tags: TxTag[], blockHash?: string): Promise<ProtocolFees[]> {
+    const args = (blockHash ? { tags, blockHash } : { tags }) as unknown as { tags: TxTag[] };
```
polymath-eric, line 34:
```suggestion
    const args = (blockHash ? { tags, blockHash } : { tags }) };
```

There shouldn't be type casts like this outside of tests.

